PR #6, opened 2015-12-08
Fix: EventId is always 0

### LINKED ISSUE #4: TraceListener adds EventId property that is always 0
When 'public override void Write(string message)' override is used EventId property is added with value 0. We should not add this property for this override. For other overrides 0 is a legitimate value. (Consider making EventId nullable and ckeck for null)

abaranch:
Fix #4


SergeyKanzhelev:
:shipit:

PR #16, opened 2016-02-06
Fix for issue that caused build to not clean

madeleinefletcher:


SergeyKanzhelev:
Just wonder how it helps build?

:shipit: anyway

madeleinefletcher:
The $(BinRoot) variable wasn't defined, so it wasn't cleaning up the bin folders. We were ending up with a huge number of old nuget packages in our build artifacts.

SergeyKanzhelev:
makes sense. Thanks for explanaition

Issue #20, opened 2016-03-25
Support NLog layout


abaranch:
Read this: http://stackoverflow.com/questions/36083778/how-do-i-specify-a-layout-for-an-applicationinsights-nlog-target

PR #24, opened 2016-04-15
Add layout support

abaranch:
#20

### DIFF
--- a/src/Adapters.Tests/NLogTarget.Tests.Shared/NLogTargetTests.cs
+++ b/src/Adapters.Tests/NLogTarget.Tests.Shared/NLogTargetTests.cs
@@ -126,6 +126,37 @@
 
         [TestMethod]
         [TestCategory("NLogTarget")]
+        public void TraceMessageCanBeFormedUsingLayout()
+        {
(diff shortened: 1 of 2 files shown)


src/Adapters.Tests/NLogTarget.Tests.Shared/NLogTargetTests.cs
```diff
@@ -126,6 +126,37 @@ public void TraceHasTimestamp()
 
         [TestMethod]
         [TestCategory("NLogTarget")]
+        public void TraceMessageCanBeFormedUsingLayout()
+        {
+            ApplicationInsightsTarget target = new ApplicationInsightsTarget();
+            target.Layout = @"${uppercase:${level}} ${message}";
+
+            Logger aiLogger = this.CreateTargetWithGivenInstrumentationKey("test", null, target);
```
dnduffy: This is cool because if you want to switch to ILogger in future there is support for NLog and other frameworks already and you are also abstracting the logger instance acquisition. :+1:

dnduffy:
:shipit:

Issue #29, opened 2016-05-03
Occasional "item with the same key has been added" exceptions in Microsoft.ApplicationInsights.NLogTarget.ApplicationInsightsTarget.BuildPropertyBag


ericjmooney:
I have noted the following exception to occur from within some web jobs (and therefore causing them to die) when a particular logging path takes place.  I'm still trying to figure out exactly what is causing this path, but I believe that the BuildPropertyBag shouldn't cause an unhandled exception that crashes my web job.
Note that this occurs *sometimes* in the midst of otherwise normal logging (including Exceptions, Traces, etc).

Thank you!

`[05/02/2016 21:14:30 > 2d9624: ERR ] Unhandled Exception: System.ArgumentException: An item with the same key has already been added.
[05/02/2016 21:14:30 > 2d9624: ERR ]    at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
[05/02/2016 21:14:30 > 2d9624: ERR ]    at System.Collections.Generic.Dictionary``2.Insert(TKey key, TValue value, Boolean add)
[05/02/2016 21:14:30 > 2d9624: ERR ]    at System.Collections.Generic.Dictionary``2.Add(TKey key, TValue value)
[05/02/2016 21:14:30 > 2d9624: ERR ]    at Microsoft.ApplicationInsights.NLogTarget.ApplicationInsightsTarget.BuildPropertyBag(LogEventInfo logEvent, ITelemetry trace)
[05/02/2016 21:14:30 > 2d9624: ERR ]    at Microsoft.ApplicationInsights.NLogTarget.ApplicationInsightsTarget.SendException(LogEventInfo logEvent)
[05/02/2016 21:14:30 > 2d9624: ERR ]    at Microsoft.ApplicationInsights.NLogTarget.ApplicationInsightsTarget.Write(LogEventInfo logEvent)
[05/02/2016 21:14:30 > 2d9624: ERR ]    at NLog.Targets.Target.Write(AsyncLogEventInfo logEvent)
[05/02/2016 21:14:30 > 2d9624: ERR ]    at NLog.Targets.Target.Write(AsyncLogEventInfo[] logEvents)
[05/02/2016 21:14:30 > 2d9624: ERR ]    at NLog.Targets.Target.WriteAsyncLogEvents(AsyncLogEventInfo[] logEvents)
[...]

PR #30, opened 2016-05-06
Potential fix for issue 29

ericjmooney:
This is a potential fix for issue 29 which I reported recently.
(https://github.com/Microsoft/ApplicationInsights-dotnet-logging/issues/29)


abaranch:
Thank you for your contribution.
Do you know if values for duplicate keys are different? Why do we want to add both? Wouldn't that be better to ignore second pair?

ericjmooney:
I don't have enough run time on this solution (yet) to have a good idea that the values for duplicate keys might be the same or different.  My goal here is to guard against throwing away potentially useful data, and since I was unable to determine where the duplicate values are coming from I thought this was the safest solution.

abaranch:
1. We can consider checking the value and if they are the same , do not add
2. Could you please add a unit test for it?

ericjmooney:
Both are good suggestions...I will implement and create a new pull request soon.
Thank you.

PR #31, opened 2016-05-09
Potential fix for issue 29

ericjmooney:
This is a potential fix for issue 29 which I reported recently, plus changes suggested by abaranch
(#29)


abaranch:
Thank you for your contribution but it is impossible to review when the whole file is changed. Could you please check your VS settings and make sure you use spaces and not tabs? Also please untablify files you've changed so we can see only relevant changes.

src/Adapters/NLogTarget.Shared/ApplicationInsightsTarget.cs
```diff
@@ -1,202 +1,212 @@
...
+                    }
+
+                    string value = valueObj.ToString();
+                    if (propertyBag.ContainsKey(key))
+                    {
+                        if (value == propertyBag[key])
+                        {
+                            continue;
+                        }
+                        key += "_1";
```
dnduffy: Just hardcoding "_1" won't deal with more than a single duplicate, what happens when "_1" is already present?

dnduffy: It is possible the file formatting issue leading to the diff problem is based on the LF vs. CR LF differences between different editors and platforms.

ericjmooney: Thanks for your comments.

I don't believe it is possible to have more than one duplicate here.  Please correct me if I'm wrong.  The reason this issue happens at all is because the code is specifically calling out a few property names (LoggerName, UserStackFrame and UserStackFrameNumber) to populate before copying from IDictionary (logEvent.Properties) to IDictionary (trace.Properties).  Or are you suggesting that this code should be prepared to handle an implementation of IDictionary that allows for duplicate keys?

RE: the diff issue...I honestly cannot figure out what went wrong there.  Pulling the original and the one I committed and comparing them w/ a hex editor for a few minutes shows that the tabs/spaces and line endings are the same.  Will spend more time on this issue shortly.

ericjmooney:
OK -- real sorry about the shenanigans here.  That was the first (and will be the last) time that I use the git integration w/ Visual Studio.  Looks like it decided to remove every line and then re-insert them...cannot even fathom why that happened.  Going to nuke this fork and try merging the changes again.

PR #32, opened 2016-05-10
Possible fix for issue 29

ericjmooney:
Possible fix for issue 29.  Now with working comparisons!

### DIFF
--- a/src/Adapters/NLogTarget.Shared/ApplicationInsightsTarget.cs
+++ b/src/Adapters/NLogTarget.Shared/ApplicationInsightsTarget.cs
@@ -153,11 +153,21 @@ namespace Microsoft.ApplicationInsights.NLogTarget
                 {
                     string key = keyValuePair.Key.ToString();
                     object valueObj = keyValuePair.Value;
-                    if (valueObj != null)
+                    if (valueObj == null)
                     {
-                        string value = valueObj.ToString();
-                        propertyBag.Add(key, value);
+                        continue;
                     }
+
+                    string value = valueObj.ToString();
+                    if (propertyBag.ContainsKey(key))
+                    {
+                        if (value == propertyBag[key])
+                        {
+                            continue;
+                        }
+                        key += "_1";
+                    }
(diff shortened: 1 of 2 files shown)


src/Adapters/NLogTarget.Shared/ApplicationInsightsTarget.cs
```diff
@@ -153,11 +153,21 @@ private void BuildPropertyBag(LogEventInfo logEvent, ITelemetry trace)
...
                     }
+
+                    string value = valueObj.ToString();
+                    if (propertyBag.ContainsKey(key))
+                    {
+                        if (value == propertyBag[key])
+                        {
+                            continue;
+                        }
+                        key += "_1";
```
dnduffy: Glad to see the diff resolved. :)
I think this would be safer if it handled _N instead of just _1.
Since the scenario of an existing key should not be the normal path I wouldn't worry about optimizing for a single duplicate hardcoded as _1 but would instead either keep a counter while looking for an unused key or just take the current number of clock ticks or something fast and unique to use as a suffix if you are concerned you will have a lot of duplicates.
The interesting question will be if you really only have one duplicate, and if so why, or if you can actually run into a situation where there are many duplicates in which case it seems there is something else to investigate as to why that is happening.

ericjmooney:
I don't believe it is possible to have more than one duplicate here. Please correct me if I'm wrong. The reason this issue happens at all is because the code is specifically calling out a few property names (LoggerName, UserStackFrame and UserStackFrameNumber) to populate before copying from IDictionary (logEvent.Properties) to IDictionary (trace.Properties). Or are you suggesting that this code should be prepared to handle an implementation of IDictionary that allows for duplicate keys?

dnduffy:
After further discussion we will accept the _1 workaround as provided.
I am still curious what is causing this intermittent behavior for you since it hasn't happened for anyone else to our knowledge.  I am wondering if you are perhaps somehow tracing/logging the same event multiple times per occurrence in some cases such that the second call runs into this situation?

PR #34, opened 2016-06-09
Change version from x.x.x.x to x.x.x-x

abaranch:


Logging.sln
```diff
@@ -1,9 +1,9 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
+# Visual Studio 2013
+VisualStudioVersion = 12.0.40629.0
```
dnduffy: Why downgrade the version of VS?

src/Adapters.Tests/Log4NetAppender.Tests.Shared/ApplicationInsightsAppenderTests.cs
```diff
@@ -1,11 +1,4 @@
-﻿// -----------------------------------------------------------------------
-// <copyright file="ApplicationInsightsAppenderTests.cs" company="Microsoft">
-// Copyright (c) Microsoft Corporation. 
-// All rights reserved.  2013
-// </copyright>
-// -----------------------------------------------------------------------
```
dnduffy: While I realize these comments aren't everywhere they do make StyleCop complain about one less thing.

Logging.sln
```diff
@@ -76,32 +76,50 @@ Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "NLogTarget.Tests.Shared", "
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "TraceListener.Tests.Shared", "src\Adapters.Tests\TraceListener.Tests.Shared\TraceListener.Tests.Shared.shproj", "{17EE52DA-5460-4BD2-A70F-529D2A94C938}"
 EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "CommonShared", "src\Adapters\CommonShared\CommonShared.shproj", "{587B624B-8C64-498E-93D7-A2D2ABC17EAB}"
```
dnduffy: Why do some of the project paths have src removed and some don't?

src/Adapters/CommonShared/SdkVersionUtils.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        internal static string GetSdkVersion(string versionPrefix)
+        {
+            string versionStr = typeof(SdkVersionUtils).Assembly.GetCustomAttributes(false)
+                    .OfType<AssemblyFileVersionAttribute>()
+                    .First()
+                    .Version;
+
+            Version version = new Version(versionStr);
+            return (versionPrefix ?? string.Empty) + version.ToString(3) + "-" + version.Revision;
```
dnduffy: Other than a minorly different implementation of this last line what is the difference between this method and GetExpectedSdkVersion earlier in the review?

Logging.sln
abaranch: I was adding sharep projects that is much easier to do in VS 2013 where I have an extension for that. And it will open in both VS.

---
In reply to: [66371203](https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/34#discussion_r66371203) [](ancestors = 66371203)

src/Adapters.Tests/Log4NetAppender.Tests.Shared/ApplicationInsightsAppenderTests.cs
abaranch: It is not complaining about this one and in Core and Web we removed all headers and the rule. But I can put it back though I think this comment is useless.

---
In reply to: [66371291](https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/34#discussion_r66371291) [](ancestors = 66371291)

Logging.sln
abaranch: I did not do that manually, that is what VS is adding and I would not want to change it since it is working as expected.

---
In reply to: [66372033](https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/34#discussion_r66372033) [](ancestors = 66372033)

src/Adapters/CommonShared/SdkVersionUtils.cs
abaranch: One is product code, second one is test code. I want them to be different so when you change product code and introduce a bug you catch it.


---
In reply to: [66372527](https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/34#discussion_r66372527) [](ancestors = 66372527)

dnduffy:
:shipit:

PR #36, opened 2016-06-13
Add trace message to custom properties #28

### LINKED ISSUE #28: Use trace message if exception message is empty
logger.Error(new ApplicationException(), "ErrorMessage")
In this case "ErrorMessage" is ignored and not sent to AI.
We can
 - Add this as exception telemetry message if exception message is empty
 - Consider adding it in properties if exception message is not empty

abaranch:
To resolve #28 : in this version we will add custom trace message to custom properties.
After we update to Core SDK 2.2 we can switch: add exception message in custom properties and trace message use as main exception telemetry message. (To allow this there was a change in Core SDK: https://github.com/Microsoft/ApplicationInsights-dotnet/pull/242 )

### DIFF
--- a/src/Adapters.Tests/NLogTarget.Tests.Shared/NLogTargetTests.cs
+++ b/src/Adapters.Tests/NLogTarget.Tests.Shared/NLogTargetTests.cs
@@ -232,8 +232,27 @@
                 aiLogger.Debug(exception, "testing exception scenario");
             }
 
-            var telemetry = (ExceptionTelemetry)this.adapterHelper.Channel.SentItems.FirstOrDefault();
+            var telemetry = (ExceptionTelemetry)this.adapterHelper.Channel.SentItems.First();
             Assert.AreEqual(expectedException.Message, telemetry.Exception.Message);
+        }
+
+        [TestMethod]
+        [TestCategory("NLogTarget")]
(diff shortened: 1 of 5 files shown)


src/Adapters.Tests/NLogTarget.Tests.Shared/NLogTargetTests.cs
```diff
@@ -232,12 +232,31 @@ public void TraceAreEnqueuedInChannelAndContainExceptionMessage()
...
 
         [TestMethod]
         [TestCategory("NLogTarget")]
+        public void CustomMessageIsAddedToExceptionTelemetryCustomProperties()
+        {
+            Logger aiLogger = this.CreateTargetWithGivenInstrumentationKey();
+            
+            try
+            {
+                throw new Exception("Test logging exception");
```
dnduffy: Do you have to throw it in a try catch or can you just create it and pass it to the aiLogger.Debug(...) call without the extra lines of code?

abaranch: We want to have a real exception with stack.

---
In reply to: [66859585](https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/36#discussion_r66859585) [](ancestors = 66859585)

dnduffy:
:shipit:

PR #40, opened 2016-06-30
Re-adding packages

dnduffy:


src/Adapters.Tests/Log4NetAppender.Net45.Tests/packages.config
```diff
@@ -2,5 +2,8 @@
 <packages>
   <package id="log4net" version="2.0.5" targetFramework="net45" />
   <package id="Microsoft.ApplicationInsights" version="2.1.0" targetFramework="net45" />
+  <package id="Microsoft.Bcl" version="1.1.10" targetFramework="net45" />
```
abaranch: Why do we need this in 4.5 project?..

abaranch:
:shipit:

src/Adapters.Tests/Log4NetAppender.Net45.Tests/packages.config
dnduffy: I don't think I do but I was putting back all of the references which had existed in develop prior to the merges I made that took away some references that shouldn't have been removed.
I can try removing these again and see if the new build still passes.

abaranch: I do not see you putting back getlink. Is that OK?

dnduffy: It was actually only in one Net40 project.

PR #54, opened 2016-09-14
Updated some package references.

dnduffy:


abaranch:
Those kind of changes need to be included in release notes. Could you please update changelog?

abaranch:
Why there are no build results?

abaranch:
Why nuspec is not updated?

dnduffy:
Updated as requested.

Issue #67, opened 2017-01-25
ApplicationInsightsTraceListener does not respect Flush


markanewman:
When calling Trace.Flush(), some of the last messages may be lost. Per [this](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-windows-desktop), TelemetryClient.Flush needs to be called to make sure the last batch is sent.

kavanj:
A similar issue exists in `ApplicationInsightsAppender` (Log4NetAppender).
`ApplicationInsightsAppender` doesn't override `AppenderSkeleton.Flush()`, so calling `Flush` just hits [AppenderSkeleton.Flush()](https://github.com/apache/log4net/blob/trunk/src/Appender/AppenderSkeleton.cs#L792).

I haven't experimented with `ApplicationInsightsTarget` (NLogTarget), but I'd expect the same issue to exist there, since there doesn't appear to be anything calling `TelemetryClient.Flush()`

markanewman:
I am not familiar with NLog, but I looked for `AppenderSkeleton.Flush()` before I put in my PR and was unable to find it. The [version](https://www.nuget.org/packages/log4net/2.0.5) that is referenced doesn't implement the `IFlushable` interface in the latest [code](http://svn.apache.org/viewvc/logging/log4net/trunk/src/Appender/AppenderSkeleton.cs?view=markup). To my eye, this seems to be the issue with #68.

PR #69, opened 2017-01-25
Issue67

markanewman:
PR for #67

### DIFF
--- a/src/Adapters.Tests/Shared/CustomTelemetryChannel.cs
+++ b/src/Adapters.Tests/Shared/CustomTelemetryChannel.cs
@@ -74,7 +74,7 @@ namespace Microsoft.ApplicationInsights
 
         public void Flush()
         {
-            throw new System.NotImplementedException();
+            throw new Exception("Flush called");
         }
 
         public void Dispose()
--- a/src/Adapters.Tests/TraceListener.Tests.Shared/ApplicationInsightsTraceListenerTests.cs
+++ b/src/Adapters.Tests/TraceListener.Tests.Shared/ApplicationInsightsTraceListenerTests.cs
@@ -367,6 +367,25 @@
             Assert.IsFalse(telemetry.Properties.ContainsKey("CallStack"));
         }
 
+        [TestMethod]
+        public void TraceListenerFlushesChannel()
+        {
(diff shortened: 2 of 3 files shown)


SergeyKanzhelev:
Would you mind making a unit test for this?

markanewman:
I did try that, but there is something wrong with my VS setup for this project. When I go to test runner, nothing happens. The build ticks off just fine, the output window reports success, but then nothing happens. I am use to the green lights showing up in test runner, but there is just nothing there.

Any ideas as to what might be up?

SergeyKanzhelev:
I haven't seen errors like this. @cijothomas, @dnduffy ?

dnduffy:
@markanewman Try exiting all instances of VS and restarting one instance and reloading the project.  I've seen this behavior before although I don't know the specifics or why it sometimes happens.  For me anyway one solution was restarting VS from scratch.  Another thing to check might be that you have the latest updates installed for VS and the latest packages installed for the unit tests.

SergeyKanzhelev:
@markanewman have you had a chance to add tests to this PR?

markanewman:
@dnduffy I also tried to reinstall VS and still no dice. My unit tests in my other projects work fine. At work we are federated with MS, should I try to contact you directly and do a screen share?

@SergeyKanzhelev If I can't get the unit tests working, should I just abandon the PR because I cant prove the work?

SergeyKanzhelev:
@markanewman this is a good feature - so please do not close PR yet. I'd ask you to try one more thing with Unit Tests before passing this PR for us to update and complete.

Did you reinstall VS on the same machines? Are other projects you mentioned uses VS tests or NUnit or xUnit? Does output window (make sure to switch to `Tests`) shows any errors? 

Here is a [long thread](http://stackoverflow.com/questions/29560011/visual-studio-2015-does-not-discover-unit-tests) on which folders could have cached bad files. Could you please try cleaning those folders.

markanewman:
@SergeyKanzhelev I reinstalled on my primary PC. I also tried to load the solution on my Surface, same thing. I us the default MSTests in VS. I will try out the PS script you mention in the link later today and report back.

markanewman:
@SergeyKanzhelev Still no dice. It just wont pick up the tests.

PR #72, opened 2017-02-16
Adding a module to support collecting ETW trace

xiaomi7732:


src/Adapters/Etw.Net451/EtwTelemetryModule.cs
```diff
@@ -0,0 +1,247 @@
...
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Disposes the module.
+        /// </summary>
+        /// <param name="isDisposing">Indicate if it is called by Dispose().</param>
+        protected virtual void Dispose(bool isDisposing)
+        {
+            if (this.isDisposed)
```
SergeyKanzhelev: why do you need this? If it's not safe to call `Dispose` twice on `traceEventSession` you need to synchronize disposing properly with `Interlocked` operations. Otherwise - remove this logic

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
```diff
@@ -0,0 +1,247 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Disposes the module.
+        /// </summary>
+        public void Dispose()
+        {
+            this.Dispose(true);
+            GC.SuppressFinalize(this);
```
SergeyKanzhelev: this class do not have finalizer. Why do you need to suppress it?

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
```diff
@@ -0,0 +1,247 @@
...
+            this.enabledProviderNames.Add(providerName);
+        }
+
+        private void DisableProviders()
+        {
+            foreach (Guid id in this.enabledProviderIds)
+            {
+                this.traceEventSession.DisableProvider(id);
+            }
+            foreach (string providerName in this.enabledProviderNames)
```
SergeyKanzhelev: stylecop will complain for closing bracket without following newline. Please run build in release as there might be more issues

src/Adapters/Etw.Net451/Implemenetation/TraceEventExtensions.cs
```diff
@@ -0,0 +1,118 @@
...
+            Debug.Assert(client != null, "Should always receive a valid client");
+
+            string formattedMessage = traceEvent.FormattedMessage;
+            TraceTelemetry telemetry = new TraceTelemetry(formattedMessage, traceEventLevelToSeverityLevel[(int)traceEvent.Level]);
+
+            telemetry.AddProperty("EventId", traceEvent.ID.ToString());
+            telemetry.AddProperty(nameof(traceEvent.EventName), traceEvent.EventName);
+            if (traceEvent.ActivityID != default(Guid))
+            {
+                telemetry.AddProperty(nameof(traceEvent.ActivityID), traceEvent.ActivityID.ToString());
```
SergeyKanzhelev: hm. we may need to think of aligning `ActivityID` and `RelatedActivityID` to the fields like `Context.Operaiton.ParentID`. I do not have exact solution right now. What do you think?

src/Adapters/Etw.Net451/Implemenetation/TraceEventExtensions.cs
```diff
@@ -0,0 +1,118 @@
...
+            }
+            if (traceEvent.RelatedActivityID != default(Guid))
+            {
+                telemetry.AddProperty(nameof(traceEvent.RelatedActivityID), traceEvent.RelatedActivityID.ToString());
+            }
+            telemetry.AddProperty(nameof(traceEvent.Channel), traceEvent.Channel.ToString());
+            telemetry.AddProperty(nameof(traceEvent.Keywords), GetHexRepresentation((long)traceEvent.Keywords));
+            telemetry.AddProperty(nameof(traceEvent.Opcode), traceEvent.Opcode.ToString());
```
SergeyKanzhelev: Similar to above - maybe we will need to recognize `Start` and `Stop` and convert those to `RequestTelemetry`

src/Adapters/EventSource.Shared/EventSource.Shared/Implementation/EventSourceListenerEventSource.cs
```diff
@@ -19,6 +19,8 @@ internal sealed class EventSourceListenerEventSource : EventSource
         public static readonly EventSourceListenerEventSource Log = new EventSourceListenerEventSource();
 
         private const int NoEventSourcesConfiguredEventId = 1;
+        private const int FailedToEnableProvidersEventId = 2;
+        private const int ModuleInitializationFailedEventId = 3;
 
         private EventSourceListenerEventSource()
```
SergeyKanzhelev: Cannot find event for "need to run under priviledged account". When you'll create it - mark it with "UserActionable" keyword and make it Error. So it will be send to portal and customer will see it

src/NuGet/EtwCollector/ApplicationInsights.config.install.xdt
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0"?>
+<ApplicationInsights xmlns="http://schemas.microsoft.com/ApplicationInsights/2013/Settings" xmlns:xdt="http://schemas.microsoft.com/XML-Document-Transform">
+  <TelemetryModules xdt:Transform="InsertIfMissing">
+    <Add Type="Microsoft.ApplicationInsights.EtwCollector.EtwTelemetryModule, Microsoft.ApplicationInsights.EtwCollector" xdt:Locator="Match(Type)" xdt:Transform="InsertIfMissing">
+      <Sources>
+        <!-- Replace/add your own ETW provider as necessary. -->
+        <Add ProviderName="Some-ETW-Provider-Name" Level="Warning"/>
```
SergeyKanzhelev: so you want a "broken" provider be registered by default? Can you include some useful provider instead? Like "process start/process stop" provider. Or have a hardcoded logic that will not try to register this provider

src/NuGet/EtwCollector/Package.nuspec
```diff
@@ -0,0 +1,36 @@
...
+  <files>
+    <!-- Binaries for .NET -->
+    <file src="$configuration$\src\Adapters\Etw.Net451\Microsoft.ApplicationInsights.EtwCollector.dll" target="lib\net451" />
+    <file src="$configuration$\src\Adapters\Etw.Net451\Microsoft.ApplicationInsights.EtwCollector.pdb" target="lib\net451" />
+    <file src="$configuration$\src\Adapters\Etw.Net451\Microsoft.ApplicationInsights.EtwCollector.xml" target="lib\net451" />
+    <!-- XDT transformations -->
+    <file src="$configuration$\src\Nuget\EtwCollector\ApplicationInsights.config.install.xdt" target="content"/>
+    <file src="$configuration$\src\Nuget\EtwCollector\ApplicationInsights.config.uninstall.xdt" target="content"/>
+    <!-- Custom targets -->
+    <file src="$configuration$\src\Nuget\EtwCollector\build\Microsoft.ApplicationInsights.EtwCollector.targets" target="build"/>
```
SergeyKanzhelev: please do not include this target. We have a very bad experience with targets when projects were not compiling after enabling of Application Insights. Or some problems with project reloading happened.

src/NuGet/EtwCollector/build/Microsoft.ApplicationInsights.EtwCollector.targets
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="_Ms-AI-EventSourceListener-EnsureApplicationInsightsConfig" BeforeTargets="PrepareForBuild">
+    <ItemGroup>
+      <_AppInsightsConfigFile Include="@(None)" Condition="'%(Filename)%(Extension)' == 'ApplicationInsights.config'" />
+    </ItemGroup>
+    <Warning Text="Can’t find ApplicationInsights.config. Without this file Application Insights may not work." Condition="@(_AppInsightsConfigFile) == ''" />
```
SergeyKanzhelev: It is a supported scenario to not have this file. No need to warn people

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
xiaomi7732: Yes, there is no finalizer and the line of SuppressFinalize is optional.
I followed basic dispose pattern here: https://msdn.microsoft.com/en-us/library/b1yfkh5e(v=vs.110).aspx#Anchor_0.

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
xiaomi7732: In this way, Dispose is allowed to be called more than once by choosing to do nothing after the first call.
I am a bit worried to use Interlocked operation because:
1. Avoid lock overhead as much as possible.
2. Do not want to mess with dispose.

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
xiaomi7732: I'll make the change. But just FYI, stylecop is not complaining. :-)

src/Adapters/EventSource.Shared/EventSource.Shared/Implementation/EventSourceListenerEventSource.cs
xiaomi7732: Sure.

src/NuGet/EtwCollector/Package.nuspec
xiaomi7732: Sure.

src/NuGet/EtwCollector/build/Microsoft.ApplicationInsights.EtwCollector.targets
xiaomi7732: Sure. Removed in the next iteration.

src/NuGet/EtwCollector/ApplicationInsights.config.install.xdt
xiaomi7732: Fixed in the next iteration.

src/Adapters/Etw.Net451/Implemenetation/TraceEventExtensions.cs
xiaomi7732: Are you asking EtwCollect trying to understand the event and then category it to different ITelemetry objects like a Trace, a Request, even an Event or something else? That sounds like an interesting issue. I think it worth having a separate issue for that discussion.

src/Adapters/Etw.Net451/Implemenetation/TraceEventExtensions.cs
xiaomi7732: Same as above: worth a different issue for tracking.

xiaomi7732:
Minor issues are fixed. Leaving event type converting out of this effort because that worth a separate issue.

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
SergeyKanzhelev: ok

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
SergeyKanzhelev: My point is you need to decide whether it's safe to call `Dispose` of a child twice or not. If it's sage - no need for this logic. If not - we have to avoid "ghost bug chasing" by introducing locks

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
SergeyKanzhelev: This is strange - maybe this repo uses a different ruleset

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
xiaomi7732: Sorry to make you state your point again because I've been mix things for my first comment. Let me try again to see if I can make it clear.

1. Your major concern is that it is not clear if Dispose is allowed to be called more than once, right?
And the answer is yes, it is safe here when Dispose(bool) is called more than once.

2. Your second concern is that the check for 'isDisposed' is not enough a guarantee of code being called more than once.
And that is correct. Since Dispose() is allowed to be called more than once concurrently, it is OK if more than one threads pass the point of the check.
One of the small benefit of having the check is outside the scenario of concurrency calling, it will save a little bit calling into traceEventSession.Dispose(). And that is it.

3. Let me know if bullet 1 or 2 is not clear or if you still think I missed some points because I want to full understand your idea before taking actions on the code.

Also let me know if you still feel uncomfortable of having the check of 'isDisposed' and I'll take it as a final decision and I'll remove it from the code.

SergeyKanzhelev: I understand the point, I do not see a benefit having the conditional logic and extra flag when we can avoid it.

xiaomi7732:
Could you please revisit the decision on the Dispose and let me know your final decision? I want to give it a last shoot because I think that's the optimized code. But I am also worried that we will miss something more important if we got distracted by those small details. If you could just make the call, I will go whatever the decision is as long as there's no defects :-) And thanks for the review!

SergeyKanzhelev:
@xiaomi7732 please remove the check and I'm ok to merge

src/Adapters/Etw.Net451/EtwListeningRequest.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Verify this request is valid.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">Throws when the object is not valid.</exception>
+        public void Validate()
+        {
+            if (this.ProviderGuid == Guid.Empty && string.IsNullOrEmpty(this.ProviderName))
+            {
+                throw new ArgumentException("ProviderGuid and ProviderName can't be null at the same time.");
```
karolz-ms: I would just return true/false with `out string` parameter for the validation error message.

src/Adapters/Etw.Net451/ITraceEventSession.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        /// <summary>
+        /// If this is a real time session you can fetch the source associated with the session
+        /// to start receiving events. Currently does not work on file based sources (we
+        /// expect you to wait until the file is complete).
+        /// </summary>
+        ETWTraceEventSource Source { get; }
+
+        /// <summary>
+        /// Is the current process Elevated (allowed to turn on a ETW provider). This is
```
karolz-ms: See the comment above. This is not technically true and in general, I do not think it makes sense to do this kind of checking in our code. Just rely on exceptions from TraceEvent library and/or the operating system

src/Adapters/Etw.Net451/ITraceEventSession.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// useful because you need to be elevated to enable providers on a TraceEventSession.
+        /// </summary>
+        bool? IsElevated();
+
+        /// <summary>
+        /// Enable a NON-KERNEL provider (see also EnableKernelProvider) which has a given
```
karolz-ms: Don't copy docs from other libraries without reading them first 😃 

Since you do not have `EnableKernelProvider`, this is just too verbose and confusing. Just say "Enables ETW provider with given provider name" or something like that.

src/Adapters/Etw.Net451/ITraceEventSession.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// provider name. This API first checks if a published provider exists by that name,
+        /// otherwise it assumes it is an EventSource and determines the provider Guid by
+        /// hashing the name according to a well known algorithm. Thus it will never return
+        /// a failure for a incorrect spelling of the name.
+        /// </summary>
+        /// <param name="providerName">
+        /// The name of the provider. It must either be registered with the operating system
```
karolz-ms: "The name of the provider" is perfectly sufficient

src/Adapters/Etw.Net451/ITraceEventSession.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <param name="providerGuid">The Guid that represents the event provider enable.</param>
+        /// <param name="providerLevel">The verbosity to turn on.</param>
+        /// <param name="matchAnyKeywords">
+        /// A bit vector representing the areas to turn on. Only the low 32 bits are used
+        /// by classic providers and passed as the 'flags' value. Zero is a special value
+        /// which is a provider defined default, which is usually 'everything'.
+        /// </param>
+        /// <param name="options">Additional options for the provider (e.g. taking a stack trace), arguments ...</param>
+        /// <returns>true if the session already existed and needed to be restarted.</returns>
+        bool EnableProvider(Guid providerGuid, TraceEventLevel providerLevel = TraceEventLevel.Verbose, ulong matchAnyKeywords = ulong.MaxValue, TraceEventProviderOptions options = null);
```
karolz-ms: Since `EtwListeningRequest` does not include `TraceEventOptions` I am not sure if it makes sense to have the options parameter on this call.

src/Adapters/Etw.Net451/Implemenetation/TraceEventExtensions.cs
karolz-ms: @SergeyKanzhelev good question that we should probably take offline. This ties to the correlation work that Liudmila is doing as well. I am more-or-less familiar with how ETW Activity IDs work but I am not sure how they tie to AI Operation IDs.

karolz-ms: @xiaomi7732 Why not use ActivityPathDecoder to decode hierarchical activity IDs?

src/Adapters/Etw.Net451/Implemenetation/TraceEventExtensions.cs
karolz-ms: @SergeyKanzhelev The problem is that opcodes are not consistently applied to events. That is why the managed activity stuff relies on event naming convention (`SomethingStart` and `SomethingStop`) do denote activity boundaries.

Also not all activities are requests. I think a better way is to do the planned metadata support work and rely on attached metadata to transform ordinary traces into requests, according to configuration describing specific events that should be treated as requests.

src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                Assert.AreEqual(1, listener.EventsReceived.Count);
+                Assert.AreEqual(ModuleInitializationFailedEventId, listener.EventsReceived[0].EventId);
+                Assert.AreEqual("Can't initialize a module that is disposed. The initialization is terminated.", listener.EventsReceived[0].Payload[1].ToString());
+            }
+        }
+
+        [TestMethod]
+        [TestCategory("EtwTelemetryModule")]
+        public void EtwTelemetryModuleInitializeFailedWhenProcessNotElevated()
```
karolz-ms: See comment s below--I do not think we should check for this

src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
```diff
@@ -0,0 +1,218 @@
...
+            else
+            {
+                configuration.TelemetryChannel = this.adapterHelper.Channel;
+            }
+            return configuration;
+        }
+
+        [TestMethod]
+        [TestCategory("EtwTelemetryModule")]
+        public void EtwTelemetryModuleDefaultConstructorExists()
```
karolz-ms: (minor) I do not think it is necessary to prefix all test method names in this class with `EtwTelemetryModule`

src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
```diff
@@ -0,0 +1,218 @@
...
+                Guid guid = Guid.Empty;
+                module.Sources.Add(new EtwListeningRequest()
+                {
+                    ProviderGuid = guid,
+                    Level = Diagnostics.Tracing.TraceEventLevel.Always
+                });
+                module.Initialize(GetTestTelemetryConfiguration());
+                Assert.AreEqual(0, traceEventSession.EnabledProviderGuids.Count);
+            }
+        }
```
karolz-ms: How about some tests that check

- whether events are actually reported
- whether calling Initialize() with different configuration changes what gets reported
- whether all ETW event properties are correctly translated to TraceTelemetry properties
- whether duplicate property names are handled correctly

TraceEvent is an abstract class so it should be possible to provide your own test implementation and source mock events from test trace session

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
xiaomi7732: Sure. Done.

xiaomi7732:
Updated.
Thanks for the review. Please merge if there's no more concerns :-)

xiaomi7732:
I need some time to response karolz-ms' review. Please hold the merge a bit.

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
```diff
@@ -0,0 +1,241 @@
...
+                {
+                    EventSourceListenerEventSource.Log.NoSourcesConfigured(moduleName: nameof(EtwTelemetryModule));
+                    return;
+                }
+
+                this.EnableProviders();
+                try
+                {
+                    // Start the trace session
+                    Task.Factory.StartNew(() => this.StartTraceEventSession(this.traceEventSession, this.client), TaskCreationOptions.LongRunning);
```
Dmitry-Matveev: If Initialize can be called twice - will the task from the first time still run in the background or will it exit due to some condition change?

src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
```diff
@@ -0,0 +1,67 @@
...
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{15FA58D1-F397-4808-831C-7AE65EE640AF}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Microsoft.ApplicationInsights.EtwCollector.Tests</RootNamespace>
+    <AssemblyName>Microsoft.ApplicationInsights.EtwCollector.Net451.Tests</AssemblyName>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
```
xiaomi7732: Could you please elaborate the concerns that 4.6 is used in the Test project? I use 4.6 is because the mocked EventSource I use for the unit tests provides richer info (mainly, the payload names) that we can do more check over the event data.

src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
```diff
@@ -0,0 +1,67 @@
...
+    <FileAlignment>512</FileAlignment>
+    <ProjectTypeGuids>{3AC096D0-A1C2-E12C-1390-A8335801FDAB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">10.0</VisualStudioVersion>
```
xiaomi7732: I don't fully understand MsBuild. Since it is there by Visual Studio and it doesn't seem break anything, I prefer to keep them unless you insist to do the house cleaning :-)

src/Adapters/Etw.Net451/AITraceEventSession.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            return this.session.EnableProvider(providerName, providerLevel, matchAnyKeywords, options);
+        }
+
+        public bool? IsElevated()
+        {
+            return TraceEventSession.IsElevated();
+        }
+
+        public bool Stop(bool noThrow = false)
```
xiaomi7732: Good catch. I'll fix it.

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
```diff
@@ -0,0 +1,246 @@
...
+            this.traceEventSession = traceEventSession;
+            this.StartTraceEventSession = startTraceEventSessionAction;
+        }
+
+        /// <summary>
+        /// Gets the list of ETW Provider listening requests (information about which providers should be traced).
+        /// </summary>
+        public IList<EtwListeningRequest> Sources { get; private set; }
+
+        private Action<ITraceEventSession, TelemetryClient> StartTraceEventSession
```
xiaomi7732: I don't see any :-) Good catch. I'll fix it.

src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
karolz-ms: In general test assembly should target the same FX version that the assembly under test targets

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
```diff
@@ -0,0 +1,246 @@
...
+        /// <summary>
+        /// Initializes the telemetry module and starts tracing ETW events specified via <see cref="Sources"/> property.
+        /// </summary>
+        /// <param name="configuration">Module configuration.</param>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            string errorMessage;
+            if (configuration == null)
+            {
+                errorMessage = string.Format(CultureInfo.InvariantCulture, "Argument {0} is required. The initialization is terminated.", nameof(configuration));
```
xiaomi7732: Ahah, I know someone is going to ask this.
Firstly, output of ToString() various on cultures that I can't use string interpolation directly. The intended way to use string interpolation with IFormatter is to make use of System.FormattableString, which, unfortunately, does not exist in .NET Fx 4.5.1.

Unless you have a better way to go, I'll have to keep using string.Format here. :-)

karolz-ms: Ah, OK, makes sense

src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
xiaomi7732: Sure.

src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
xiaomi7732: Agree. I'll fix it in the next push.

src/Adapters/Etw.Net451/EtwListeningRequest.cs
xiaomi7732: Sure.

src/Adapters/Etw.Net451/ITraceEventSession.cs
xiaomi7732: Sure.

src/Adapters/Etw.Net451/EtwTelemetryModule.cs
xiaomi7732: @Dmitry-Matveev , good good question. I intention is to keep 1 task for the session for one instance, no matter how many times initialize is called. The implementation, though, will cause multiple tasks being started.

I think there are 2 ways to fix the issue. 1 is to close the task when initialize is called using a cancellation token; 2 is to keep the session but to enable / disable different providers on the session according the change of the configuration.

I think it the best performance + as much events being captured as possible to not interrupt the session when configure changes.

So, I'll update the code to start the session only when isInitialized flag is false.

src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
xiaomi7732: Sure.

src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
xiaomi7732: That is generally true and there's usually no reason to having a higher version of target framework project to test lower version of target. In this case, it is about having a set of properties being detected verified by unit tests. Considering it is a real / possible scenario for end user, I prefer keep it this way. Let me know if you feel strongly disagree.

src/Adapters/Etw.Net451/ITraceEventSession.cs
xiaomi7732: OK.

src/Adapters/Etw.Net451/ITraceEventSession.cs
xiaomi7732: Sure.

src/Adapters/Etw.Net451/ITraceEventSession.cs
xiaomi7732: That is really detailed examination :-) I'll just keep it there since it's on the original interface and in theory, there is still a small chance that we are going to implement it later. Let me know if you don't agree :-)

xiaomi7732:
@karolz-ms / @Dmitry-Matveev ,
Please review the last iteration for the changes. Thanks!

src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
karolz-ms: No, that is fine, thanks for explanation.

src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
karolz-ms: The more properties in the project, the more opportunities to screw up or run into problems with future versions of the tooling. 

If you do not fully understand the purpose of something in the project file, I would like you to do a quick experiment: try removing the property and see if a _clean_ build still succeeds. If it does, the property is probably not necessary. If not, you just gained some understanding 😃 

In this particular case I cannot think of a good reason why this project would be limited to a single VS version, so this setting seems suspect, undesirable "VS noise"

src/Adapters/Etw.Net451/ITraceEventSession.cs
karolz-ms: There is also a larger chance that we won't need it 😄 

But I am fine with keeping this on an internal interface.

karolz-ms:
(but see if you can address my concerns about properties in the project file)

src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
xiaomi7732: I think you are over-concerning project file properties being broken. If Visual Studio is not handling the properties it adds to the project, it is a visual studio bug.

That said, I will remove it as I am not seeing a real problem either to remove the property.

On the side, what the code say is: If there is no variable of VisualStudioVersion specified, it will set it to 10.0 as a default value, which I guess is a backward compatibility safer. :-)

src/Adapters/Etw.Net451/ITraceEventSession.cs
xiaomi7732: That I agree on the larger chance part.

xiaomi7732:
Thanks @SergeyKanzhelev , @karolz-ms & @Dmitry-Matveev for your review.
@SergeyKanzhelev, please merge the change if there is more concerns.

PR #73, opened 2017-02-21
Fix unit tests failure due to racing condition

xiaomi7732:
The tests passed on the local debugging session but failed during
running. There are 2 issues contribute to the failure on the sever.

1. Initialize() will return slightly earlier than
traceEventSession.Process() is fully running. Which leads to the event
lost for the unit test because the unit test invokes the event
immediately after calling the initilize() on the module.
Although the right fix would be locking on initilize() until Process()
is fully up, it is an overhead to implement. At the same time, in real
world scenario, events are emitted in different process and will be
capatured once Process() is up.
So, I am fixing the issue by simple put a delay in the unit test after
calling initialize().

2. When using F5 (Debug) to invoke unit tests and break the test case in
the middle, the trace event session will be left. That's why after
Debug, the unit tests start to pass because there are existing opening
sessions writing the events into the channel. This introduces huge
inconsistence for the unit test result.
I added clean up logic for the tests to solve this issue.

### DIFF
--- a/src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
+++ b/src/Adapters.Tests/Etw.Net451.Tests/Etw.Net451.Tests.csproj
@@ -33,6 +33,7 @@
     <Reference Include="System" />
     <Reference Include="System.Net.Http" />
     <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework" />
+    <Reference Include="System.Security" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="EtwTelemetryModuleTests.cs" />
--- a/src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
+++ b/src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
@@ -10,6 +10,7 @@ namespace Microsoft.ApplicationInsights.EtwTelemetryCollector.Tests
     using System.Collections.Generic;
     using System.Diagnostics.Tracing;
     using System.Linq;
+    using System.Security.Principal;
     using System.Threading.Tasks;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.EtwCollector;
@@ -27,6 +28,7 @@ namespace Microsoft.ApplicationInsights.EtwTelemetryCollector.Tests
         private const int AccessDeniedEventId = 4;
 
         private readonly AdapterHelper adapterHelper = new AdapterHelper();
+        private static bool isTestEnvGood;
 
         public void Dispose()
         {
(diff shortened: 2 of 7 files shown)


karolz-ms:
Actually, I suggest that the test fix is incorrect _and_ that this failure uncovered a real issue product issue.

We should not include an arbitrary delay into unit tests.

1. It makes the test unreliable (what if the delay is still insufficient under some circumstances)?
2. It makes the test slower than it could otherwise be. 

I had exactly the same problem to solve in the EventFlow implementation. Take a look (the test event session has an event that gets set when Process() gets to run and the test waits on that event before proceeding).

The product issue is that a real TraceEventSession gets created when the Etw telemetry module gets created. That is incorrect and I missed that during code review, sorry. The trace event session creation should be delayed until Initialize().

Also, unit tests should not cause the Etw telemetry module to create a real trace event session, ever. As you discovered, this might cause undesired side effects. Best practice is to ensure maximum isolation for the code under test and a mock trace event session provides that. HTH!

xiaomi7732:
@karolz-ms, thanks for the review.
There are several points in your review and I'll explain one by another.

1. **We should not include an arbitrary delay into unit tests** - **I agree.** The reasons are good. I have addressed all of them **except one.** The only one exception is for the ReactToConfigurationChange test, where after configuration is updated, a small delay needs to be there before the pipeline start to get the events again. That I can't handle by waiting on the channel because the lost event will never be there.

At the same time, **I am not too concerns about it.** Because in the production environment, the events are provided by other processes. There needs no sync signal between the processes. Before session.Process() fully functional, no events are expected to be caught while as long as the session.Process() starts to fully function, events will then be caught.

2. We should delay the creation of the TraceEventSession until Initialize() is called - **Good point**. Sorry I didn't consider that before. It is fixed.

3. Unit tests should not cause the Etw telemetry module to create a real trace event session - Generally, I would agree this. However, ETW trace is an exception here. If we need to test the conversion of the event data object or to test the event handling against the EtwTelemetryModule, without creating a real etw session, there have to have a way to create a mock of TraceEvent. After all, that's the source data structure that we are interested in to verify against the result data (ITelemetry).

Directly mocking TraceEvent is not fit per current implementation because TraceEvent is not inheritable, nor of the existing derived classes has the ability that allow us to mock the Level property or the Message property. Unfortunately, those are the properties that we are interested in.

One way out would be abstract TraceEvent and having a wrapper for it and having all the access to use the wrapper. Technically, it is doable. But this means an additional wrapper in production code, and the only usage of the wrapper is for the unit tests. I don't think we should do that.

On the other hand, I do see the concerns that not able to creating ETW sessions because there's limitations on the user account / group / process privilege. I updated the unit test to making sure we have a good enough environment. Otherwise, the all the etw collector test cases will be skipped with a  warning of inconclusive assertion.

Hope this addresses all the concerns.

Thanks again for the review.

SergeyKanzhelev:
@xiaomi7732 do you want to follow @karolz-ms advice or this PR is good to merge?

xiaomi7732:
@karolz-ms Updated.
@SergeyKanzhelev, please help merge the change. Thank you!

PR #76, opened 2017-03-09
Adding a constructor to the TraceListener which allows the user to specify a TelemetryClient

palenshus:
This is useful if they'd like to configure their SessionId for example. Added a new UT for the change, and verified that all tests pass

### DIFF
--- a/src/Adapters/TraceListener.Shared/ApplicationInsightsTraceListener.cs
+++ b/src/Adapters/TraceListener.Shared/ApplicationInsightsTraceListener.cs
@@ -47,6 +47,16 @@ namespace Microsoft.ApplicationInsights.TraceListener
             this.TelemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("sd:");
         }
 
+        /// <summary>
+        /// Initializes a new instance of the ApplicationInsightsTraceListener class, specifying a TelemetryClient to use
+        /// </summary>
+        /// <param name="telemetryClient">TelemetryClient to use for tracing</param>
+        public ApplicationInsightsTraceListener(TelemetryClient telemetryClient)
+        {
+            this.TelemetryClient = telemetryClient;
+            this.TelemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("sd:");
+        }
+
         internal TelemetryClient TelemetryClient { get; set; }
         
         /// <summary>
--- a/src/Adapters.Tests/TraceListener.Tests.Shared/ApplicationInsightsTraceListenerTests.cs
+++ b/src/Adapters.Tests/TraceListener.Tests.Shared/ApplicationInsightsTraceListenerTests.cs
@@ -26,7 +26,7 @@
         [TestCategory("TraceListener")]
         public void TraceListenerInitializeDoesNotThrowWhenInstrumentationKeyIsNull()
         {
-            var listener = new ApplicationInsightsTraceListener(null);
+            var listener = new ApplicationInsightsTraceListener(instrumentationKey: null);
             listener.Dispose();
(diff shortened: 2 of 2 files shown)


src/Adapters/TraceListener.Shared/ApplicationInsightsTraceListener.cs
```diff
@@ -47,6 +47,16 @@ public ApplicationInsightsTraceListener(string instrumentationKey)
             this.TelemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("sd:");
         }
 
+        /// <summary>
+        /// Initializes a new instance of the ApplicationInsightsTraceListener class, specifying a TelemetryClient to use
+        /// </summary>
+        /// <param name="telemetryClient">TelemetryClient to use for tracing</param>
+        public ApplicationInsightsTraceListener(TelemetryClient telemetryClient)
```
SergeyKanzhelev: We typically initialize modules with `TelemetryConfiguration` like [this](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs#L76) Will this work for your scenario? Or you need `TelemetryClient` specifically?

palenshus: Thanks for the reply Sergey! I'm not sure if that would work, my goal is to be able to change the Session.Id during execution. With my change I could do that by modifying the telemetryClient that I'd previously passed into the TraceListener constructor.

I don't see any way to set the session with the TelemetryConfiguration class, so I don't *think* that'd work for me. Please correct me if I'm wrong

SergeyKanzhelev: Typical way doing it will be to code it like this:

``` csharp
class SessionIdTelemetryInitializer : ITelemetryInitializer
{
  public void Process(ITelemetry t)
  {
    t.Context.Session.Id = GetSessionIdForCurrentThread();
  }
}

var config = new TelemetryConfiguraiton();
config.TelemetryInitializers.Add(new SessionIdTelemetryInitializer());

var listener = new ApplicationInsightsTraceListener(config);
```

This way you can control session ID dynamically and do not need to synchronize threads.

Will this satisfy your scenario?

palenshus: Yes I believe that would work. I'm confused however, in the documentation  (https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-filtering-sampling#add-properties) it says that ITelemetryInitializer specifies the Initialize() method while ITelemetryProcessor provides the Process() method. Did you mean ITelemetryProcessor in the above example?

And just to be clear, are you proposing a new constructor that takes a TelemetryConfiguration? Would you guys be doing that work or would be fine with me adding it? Thanks!

dnduffy: We appreciate all contributions so please feel free to do it. :)
Thank-you Union.

SergeyKanzhelev: Sorry, I meant `Initialize`, not `Process`.

palenshus: Hmm, it's complaining that the TelemetryConfiguration constructor parameter is not CLS compliant. I assume that's important to you guys :) Thoughts?

MS-TimothyMothra:
I'm closing this PR because we want to take this in a different direction. 
TraceListener should accept TelemetryConfiguration in a constructor.
I've created Issue #157 to track this.

PR #77, opened 2017-03-30
Make EventSource telemetry module target netstandard1.3

karolz-ms:
This change also makes the whole AI-logging repo use MSBuild 15 tooling


Common.props
```diff
@@ -1,11 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'Global.props'))\Global.props" />
-  <Import Project=".\GlobalStaticVersion.props" />
```
SergeyKanzhelev: this change makes this repository inconsistent with the https://github.com/Microsoft/ApplicationInsights-dotnet Do you think we need to change that one too?

karolz-ms: I think that might be necessary.

The issue is that if the property is set after Microsoft.Common.props has been imported, then it is too late for it to pick up the .props files generated in that folder by NuGet restore.  You can resolve this by always importing Global.props file before Microsoft.Common.props gets imported, but I found it tricky to ensure. An easier way, and this is what I have done here, is to use the special Directory.Build.props file which will automatically get imported beforehand by MSBuild.  For more information, see https://github.com/Microsoft/msbuild/issues/1603#issuecomment-283104367

karolz-ms:
Just to clarify: this change will require that the repo is built using a machine pool with MSBuild 15/VS 2017 tooling

Common.props
SergeyKanzhelev: Is this the same idea? https://github.com/Microsoft/ApplicationInsights-dotnet/pull/466/files Can you please align by names of files

karolz-ms: It is the same problem but alternative solution. I like this one better because it does not require non-default project structure (i.e. you can say `<Project Sdk="Microsoft.NET.Sdk">` and everything still works.

Not sure exactly what you mean "align by names of files"? Do you mean "make it consistent between the two repos"?

SergeyKanzhelev: Yes. I didn't dig into the exact solution details. So if you think your solution better - can you please as a minimum create an issue in `dotnet` repo

karolz-ms: Sounds good, https://github.com/Microsoft/ApplicationInsights-dotnet/issues/490 logged

PR #78, opened 2017-03-30
Add dotnet MyGet package source

karolz-ms:
This should hopefully fix the restore issue we have seen with the last build attempt

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -19,6 +19,7 @@
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
     <add key="buildtools" value="https://www.myget.org/F/dotnet-buildtools/" />
     <add key="MyGetApplicationInsights" value="https://www.myget.org/F/applicationinsights" />
+    <add key="dotnet" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" protocolVersion="3" />
   </packageSources>
   <disabledPackageSources>
     <add key="buildtools" value="true" />


NuGet.Config
```diff
@@ -19,6 +19,7 @@
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
     <add key="buildtools" value="https://www.myget.org/F/dotnet-buildtools/" />
     <add key="MyGetApplicationInsights" value="https://www.myget.org/F/applicationinsights" />
+    <add key="dotnet" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" protocolVersion="3" />
```
SergeyKanzhelev: why do we need this? Can we limit the use to only use officially released packages?

karolz-ms: I am getting the following error

`Unable to resolve 'System.Diagnostics.DiagnosticSource (>= 4.4.0)' for '.NETStandard,Version=v1.3'`

DiagnosticSource ver. 4.4 is not on official NuGet yet. I believe this is related to the new correlation/automatic dependency tracking and brought onto us by the new Microsoft.ApplicationInsights.dll

karolz-ms: This error happens during NuGet package restoration; I am not depending on DiagnosticSource 4.4 in my code

SergeyKanzhelev: Oh, the new projects picks up the latest Application Insights NuGet from MyGet and it already has @lmolkova changes. Can you please set the max version for the dependency to be 2.3, not 2.4?

karolz-ms: Well, but I do want 2.4 because I want my EventSource telemetry module to target netstandard1.3. AFAIK the change for MS.AI.dll to target netstandard1.3 only went into 2.4 branch?

SergeyKanzhelev: Ok. Can you please create an issue similar to this https://github.com/Microsoft/ApplicationInsights-dotnet/issues/487 so we will not forget to fix it before 2.4 release

PR #84, opened 2017-05-02
Suppress TPL events for EventSource and ETW modules

karolz-ms:
The problem was found by EventFlow users, this change is a port of the fix from that library.

Basically, we want the TPL EventSource to be enabled as it provides hierarchical activity IDs. But the events emitted from that EventSource are not very useful, at least not during production tracing, so we want them disabled. Examples of events emitted by this source are "task X started task Y", "task X resumed after waiting on tasks Y and Z" and similar.

We could consider disabling the events only if they are not explicitly requested by the telemetry module configuration, but I doubt users will be interested in TPL events in the APM context.

### DIFF
--- a/src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
+++ b/src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
@@ -426,6 +426,43 @@ namespace Microsoft.ApplicationInsights.EtwTelemetryCollector.Tests
             }
         }
 
+        [TestMethod]
+        [TestCategory("EtwTelemetryModule")]
+        public async Task DoNotReportTplEvents()
+        {
+            using (var module = new EtwTelemetryModule())
+            {
+                module.Initialize(GetTestTelemetryConfiguration());
+
+                for (int i = 0; i < 10; i += 2)
+                {
+                    Parallel.For(0, 2, (idx) =>
+                    {
+                        PerformActivityAsync(i + idx).GetAwaiter().GetResult();
+                    });
+
+                }
+
+                // Wait 2 seconds to see if any events arrive asynchronously through the ETW module.
+                // This is a long time but unfortunately there is no good way to make ETW infrastructure "go faster"
+                // and we want to make sure that no TPL events sneak through.
+                await this.WaitForEventsArrive(this.adapterHelper.Channel, 1, TimeSpan.FromSeconds(2));
+
+                Assert.AreEqual(0, this.adapterHelper.Channel.SentItems.Length);
+            }
+        }
+
+        private async Task PerformActivityAsync(int requestId)
+        {
+            await Task.Run(async () =>
+            {
+                TestProvider.Log.RequestStart(requestId);
+                await Task.Delay(50).ConfigureAwait(false);
+                TestProvider.Log.RequestStop(requestId);
+
+            });
+        }
+
         /// <summary>
         /// Wait until SendItem is hit for given times or timeout in <see cref="CustomTelemetryChannel" />.
         /// </summary>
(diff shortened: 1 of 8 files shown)


src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -154,9 +154,9 @@ protected override void OnEventSourceCreated(EventSource eventSource)
         private void EnableAsNecessary(EventSource eventSource)
         {
             // Special case: enable TPL activity flow for better tracing of nested activities.
-            if (eventSource.Guid == EventSourceTelemetryModule.TplEventSourceGuid)
+            if (eventSource.Guid == TplActivities.TplEventSourceGuid)
             {
-                this.EnableEvents(eventSource, EventLevel.LogAlways, (EventKeywords)EventSourceTelemetryModule.TaskFlowActivityIds);
+                this.EnableEvents(eventSource, EventLevel.LogAlways, (EventKeywords) TplActivities.TaskFlowActivityIdsKeyword);
```
SergeyKanzhelev: I heard there is a problem with hierarchical ids on framework 46 in eventsource. Are you sure enabling those will make any good for event source listener?

karolz-ms: @SergeyKanzhelev  What is the nature of the problem? 

Enabling the TPL EventSource is definitively valuable because of the hierarchical activity ID benefit. It is a great help in figuring out how much time each step of a complicated activity took (say handling of a request that involves several async calls to dependencies)

SergeyKanzhelev: @vancem I believe somebody told me that in EventSource you wouldn't see hierarchical IDs. Was it fixed recently or still the case?

karolz-ms: @vancem @SergeyKanzhelev Event if net46 has a problem, this telemetry module targets netstandard 1.3, so it is useful for .NET Core apps too

vancem: EventSource works with Hierarchical IDs.     There was a problem where it only worked with ETW and not EventListeners (as used here), however that has been fixed in .NET Core (bug still pending to port it to Desktop).   It does work on Core

karolz-ms:
@SergeyKanzhelev Tests added

src/Adapters.Tests/Etw.Net451.Tests/TestProvider.cs
```diff
@@ -45,6 +47,18 @@ public void Tricky(int EventId, string EventName, string Message)
             WriteEvent(TrickyEventId, EventId, EventName, Message);
         }
 
+                [Event(RequestStartEventId, Level = EventLevel.Informational, ActivityOptions = EventActivityOptions.Recursive)]
```
SergeyKanzhelev: too many whitespaces

src/Adapters.Tests/Etw.Net451.Tests/EtwTelemetryModuleTests.cs
```diff
@@ -426,6 +426,43 @@ public void ProviderNotEnabledByEmptyGuid()
...
+                    });
+
+                }
+
+                // Wait 2 seconds to see if any events arrive asynchronously through the ETW module.
+                // This is a long time but unfortunately there is no good way to make ETW infrastructure "go faster"
+                // and we want to make sure that no TPL events sneak through.
+                await this.WaitForEventsArrive(this.adapterHelper.Channel, 1, TimeSpan.FromSeconds(2));
+
+                Assert.AreEqual(0, this.adapterHelper.Channel.SentItems.Length);
```
SergeyKanzhelev: can you please also test that the ID in fact will be hierarchical (meaning that Tpl was enabled)

karolz-ms:
@SergeyKanzhelev  test for hierarchical activity ID added. This was good call to add it -- turns out we did not enable the TPL provider by default for the ETW module. Thanks!

src/Adapters.Tests/Etw.Net451.Tests/TestProvider.cs
karolz-ms: fixed

PR #98, opened 2017-05-31
Reference v2.3 of AI base SDK in DiagnosticSourceListener

### LINKED ISSUE #97: DiagnosticSourceListener fails to load
The version of Microsoft.ApplicationInsights.DiagnosticSourceListener on nuget.org (2.4.0-beta2) references Microsoft.ApplicationInsights 2.4
However, the .nuspec depends on 2.3
So, we have a mismatch which causes a `System.IO.FileLoadException`, resulting in the telemetry module being rejected:

`AI: ApplicationInsights configuration file loading failed. Type 'Microsoft.ApplicationInsights.DiagnosticSourceListener.DiagnosticSourceTelemetryModule, Microsoft.ApplicationInsights.DiagnosticSourceListener' was not found. Type loading was skipped. Monitoring will continue.`

pharring:
Fixes #97

### DIFF
--- a/src/Adapters.Tests/DiagnosticSource.Tests/DiagnosticSource.Tests.csproj
+++ b/src/Adapters.Tests/DiagnosticSource.Tests/DiagnosticSource.Tests.csproj
@@ -12,7 +12,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <PackageReference Include="MSTest.TestAdapter" Version="1.1.13" />
     <PackageReference Include="MSTest.TestFramework" Version="1.1.13" />
-    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.4.0-beta3" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.3" />
     <PackageReference Include="Microbuild.Core" Version="0.2.0" />
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


cijothomas:
For all nuspecs, base sdk version is populated from the variable $coresdkversion$,  except DiagnosticSource which is hard coded to be 2.3.
https://github.com/Microsoft/ApplicationInsights-dotnet-logging/blob/develop/src/NuGet/DiagnosticSourceListener/Package.nuspec#L20

@SergeyKanzhelev @dnduffy to comment if there are any concerns with this.

cijothomas:
@pharring - if the same app uses DiagnosticSourceListener (2.3 base sdk version) and some other logging adaptor (2.4.0-beta3), there would be conflicts?

pharring:
In that case, a binding redirect will be generated.

dnduffy:
@pharring I recall a discussion about wanting to not require a beta reference, however, for 2.4.0 stable are you okay with us keeping all the versions consistent and referencing 2.4.0 of the base SDK?  We try to keep these versions in sync.

pharring:
@dnduffy No, I'm not OK with that. Requiring 2.4 limits its utility. Here's why:
Customers of Snapshot Debugging are using the base AI 2.3 SDK today. The Snapshot Collector uses DiagnosticSource for its internal logs and I want to tell customers to use the DiagnosticSourceListener to troubleshoot problems. They should not have to upgrade their app to base AI 2.4 just to get additional logging.

pharring:
@dnduffy @cijothomas would you please review?

dnduffy:
@cijothomas Are you planning to include a build of logging for 2.4.0 stable as well?

pharring:
@dnduffy @cijothomas could one of you merge, please? (I don't have permission)

Issue #122, opened 2017-11-08
Adding support for disabledEventSourceNamePrefix configuration


xiaomi7732:
Due to a .NET Framework bug: https://github.com/dotnet/coreclr/issues/14434, there's a discussion providing support for configurations to allow totally disabling some event sources by their name prefixes.
Details of the discussion could be found here: https://github.com/Azure/diagnostics-eventflow/issues/140.

@SergeyKanzhelev, @dnduffy, I am planning to put the implementation into `EventSourceTelemetryModule`.

**Here's a simple description:**
_Configuration:_
![image](https://user-images.githubusercontent.com/3674549/32577525-19a5e5aa-c48f-11e7-8b61-c7cd72a897ff.png)

_Details:_

- Disabling EventSources is not recommended under normal circumstances, however, due to some edge cases, like the bug here: https://github.com/dotnet/coreclr/issues/14434, there needs a way to work-around it.

- For event-source that matches in the 'Sources' exactly and at the same time, starts with disabledEventSourceNamePrefix, the enabling of the source will take the priority because:

  - Exact match should take over partial match;
  - Generally, for telemetry module settings, we are using the opt-in than opt-out;
  - It is flexible that user can express: I want `abc` event source but not anything else that starts with `ab`, because either I don't care or they are having issues.

Please use this issue for feedback on the ideas and I'll submit PR for review when it is implemented.

SergeyKanzhelev:
Can configuration be like this so it will be a single collection? And it will support wildcards for collecting as well.

```
<Sources>
    <Add Name = "MyCompany" Level="Verbose" Include = "true" Wildcard = "false" />
    <Add Name = "Microsoft.ApplicationInsights" Include = "false" Wildcard = "true" />
</Sources>
```

xiaomi7732:
@SergeyKanzhelev,
Thanks for the reply, especially, providing yet another implementation proposal.

Is there any specific reason you want to put them together? Performance gains?

I am leaning toward adding another section because:
1. Having disabling source with level/keywords settings on the rules feels overwhelming;
2. Using 'Add' element inside 'Sources' to prevent some event source events being gathered feels somewhat weird - I know logically, the add is for adding a rule rather than adding a source, but it just feels confusion.

At the same time, I don't see it expensive to support wildcard for opted-in event sources. I can put in the implementations together:

```xml
<Sources>
    <Add Name="MyCompany" Level="Verbose" Wildcard="false" />
[...]

PR #124, opened 2017-11-09
Add wildcards support and the ability to disable event source for AI event source listener

xiaomi7732:
Per discussion in #122, update the code with new implementations.


src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -218,5 +230,66 @@ private void EnableAsNecessary(EventSource eventSource)
...
+        /// <returns></returns>
+        internal bool IsEventSourceNameMatch(string eventSourceName, string rule)
+        {
+            if (string.IsNullOrEmpty(eventSourceName) || string.IsNullOrEmpty(rule))
+            {
+                return false;
+            }
+
+            string pattern = rule.WildCardToRegex();
+            return Regex.IsMatch(eventSourceName, pattern);
```
karolz-ms: I am concerned about the cost of regex-matching for every event. This what will happen if 10s of thousands of events need to be processed. Even if compiled regular expressions are used (which you do not use here), it is still too costly IMO.

I suggest we switch to using a simple prefix and change the disablement test to string.StartsWith()

xiaomi7732: I think this is a valid concern. I did some quick benchmark:
100,000,000 iterations on match 2 strings - 1 success, 2 fail. And here are the results:

|Method|Stopwatcher Elapsed|
|--|--|
|StartsWith| ~1.27s|
|Contains| ~ 4.7s |
|RegEx| ~ 28.21s |

Based on the result, I think we should refactor the code to use 'StartWith' for partial match.

The configuration shall look like this:
```xml
<Sources>
    <Add Name="MyCompany" Level="Verbose" IsPrefixMatch="false" />
    <Add Name="Microsoft.ApplicationInsights" IsPrefixMatch="true" />
</Sources>
<DisabledSources>
    <Add Name="SomeBadEventSource" IsPrefixMatch="true" />
</DisabledSources>
```

What do you think? @karolz-ms, @SergeyKanzhelev ?

Benchmark code just FYI:
```csharp
using System;
using System.Diagnostics;
using System.Text.RegularExpressions;

namespace RegexVSStartsWithBenchMark
{
    class Program
    {
        private Regex _regex;

        public Program()
        {
            _regex = new Regex("^hello.*", RegexOptions.Compiled);
        }

        static void Main(string[] args)
        {
            const int iterationNumber = 100000000;
            Program p = new Program();
            Stopwatch stopwatch = new Stopwatch();

            Predicate<string> predicate = p.RegExStartWith;
            // Predicate<string> predicate = p.IsStartWith;
            // Predicate<string> predicate = p.Contians;


            stopwatch.Start();
            for (int i = 0; i < iterationNumber; i++)
            {
                predicate("hello");
                predicate("xxxxx");
            }
            stopwatch.Stop();
            Console.WriteLine($"Elapsed: {stopwatch.Elapsed}");
        }
        private bool IsStartWith(string value)
        {
            return value.StartsWith("hello", StringComparison.Ordinal);
        }

        private bool RegExStartWith(string value)
        {
            return _regex.IsMatch(value);
        }

        private bool Contians(string value) {
            return value.Contains("el");
        }
    }
}
```

xiaomi7732: I've updated the implementation to use prefix match for both listening request and disabling requests so that the configuration feels more consistent.
Add sources becomes less flexible than the last iteration but at the same time, at this point, I haven't yet been able to justify that we will have a huge amount of users want using either prefix match or wildcards match for the event source names to enable.

src/Adapters/EventSource.Netstandard13/Implementation/StringExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System.Text.RegularExpressions;
+
+namespace Microsoft.ApplicationInsights.EventSourceListener.Implementation
+{
+    internal static class StringExtensions
+    {
+        public static string WildCardToRegex(this string value)
```
SergeyKanzhelev: you do not need it any longer, do you?

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -218,5 +229,65 @@ private void EnableAsNecessary(EventSource eventSource)
...
+        private bool IsDroppingEvent(EventWrittenEventArgs eventData)
+        {
+            string eventSourceName = eventData?.EventSource?.Name;
+            if (string.IsNullOrEmpty(eventSourceName))
+            {
+                throw new ArgumentNullException(nameof(eventSourceName));
+            }
+
+            // Do NOT drop the event when it matches one of the request by name exactly.
+            if (Sources.Any(request => string.Equals(request.Name, eventSourceName, StringComparison.Ordinal)))
```
SergeyKanzhelev: so you check for equality first and then iterate over the list again? I think you can either split into two lists or simply call `IsEventSourceNameMatch` in one go.

Especially keep in mind that the default config only has one namespace which is never be match exactly

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -218,5 +229,65 @@ private void EnableAsNecessary(EventSource eventSource)
...
+        /// When the event comes from a EventSource that matches the rule of DisabledSource, it should be dropped, unless the event source name matches one of the 
+        /// name exactly in the enabling rule.
+        /// </summary>
+        /// <param name="eventData">The event data to test.</param>
+        private bool IsDroppingEvent(EventWrittenEventArgs eventData)
+        {
+            string eventSourceName = eventData?.EventSource?.Name;
+            if (string.IsNullOrEmpty(eventSourceName))
+            {
+                throw new ArgumentNullException(nameof(eventSourceName));
```
SergeyKanzhelev: please do not throw

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -218,5 +229,65 @@ private void EnableAsNecessary(EventSource eventSource)
...
+
+        /// <summary>
+        /// Returns true when eventSourceName satisfied the rule; false otherwise. Returns false when either is null.
+        /// </summary>
+        /// <param name="eventSourceName"></param>
+        /// <param name="rule"></param>
+        /// <returns></returns>
+        internal bool IsEventSourceNameMatch(string eventSourceName, string rule)
+        {
+            if (string.IsNullOrEmpty(eventSourceName) || string.IsNullOrEmpty(rule))
```
SergeyKanzhelev: you've added two conditions to check that was just checked before. Why not simply call `eventSourceName.StartsWith(rule)` instead of defining whole new method?

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -218,5 +229,65 @@ private void EnableAsNecessary(EventSource eventSource)
...
+            {
+                return false;
+            }
+
+            return eventSourceName.StartsWith(rule);
```
SergeyKanzhelev: you need to specify culture for comparison. I'm surprised it wouldn't fail compilation

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -130,7 +137,11 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
             {
                 try
                 {
-                    this.onEventWrittenHandler(eventData, this.client);
+                    // We can't deal with disabled sources until event is raised due to: https://github.com/dotnet/coreclr/issues/14434
+                    if (DisabledSources.Count == 0 || !IsDroppingEvent(eventData))
```
SergeyKanzhelev: why not move the first condition inside the method?

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -218,5 +229,65 @@ private void EnableAsNecessary(EventSource eventSource)
...
+
+            return eventSourceName.StartsWith(rule);
+        }
+
+        /// <summary>
+        /// When the event comes from a EventSource that matches the rule of DisabledSource, it should be dropped, unless the event source name matches one of the 
+        /// name exactly in the enabling rule.
+        /// </summary>
+        /// <param name="eventData">The event data to test.</param>
+        private bool IsDroppingEvent(EventWrittenEventArgs eventData)
```
SergeyKanzhelev: I wonder would it be faster to keep a cache of event sources with the sampling decision instead of comparing the name all the time? Not a big deal though

src/Adapters/EventSource.Netstandard13/EventSourceListeningRequest.cs
```diff
@@ -56,7 +51,7 @@ public override bool Equals(object obj)
                 return false;
             }
 
-            return this.Name == other.Name && this.Level == other.Level && this.Keywords == other.Keywords;
+            return this.Name == other.Name && this.Level == other.Level && this.Keywords == other.Keywords && this.PrefixMatch == other.PrefixMatch;
```
SergeyKanzhelev: should you define `Equals` for base class? If so - call it's `Equals` first.

If you do not need `Equals` for base class, this `Equals` is also unnecessary

src/Adapters/EventSource.Netstandard13/Implementation/StringExtensions.cs
xiaomi7732: No, it is not needed. I'll remove the class.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: Hey, Sorry, I am not sure I got what you mean here. But I think the concern here's performance.

Firstly, in the caller, this logic will be bypassed if there's no item in <DisabledSources>.
Secondly, in the default configuration, since there are no <DisabledSources>, we won't run into this code path.

Based on those conditions, here it is checking against two different lists in terms of <Sources> or <DisabledSources>. Here's the step:
Whenever an event comes in:
1. Check if the event's source belongs to the <Sources> list. If there is a **exact** match, do NOT drop the event - as the exact match in <Sources> always take priority.

2. Otherwise, see if the event's source matches the rule in <DisabledSources>, **exact or prefix match** by the rule. If there is a match, drop the event, else, keep it.

Does this make sense? Or do I missed something? Thanks!

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: I will change this to an event source event. Thanks!

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: Yup. That is a good point. It is simple enough to call starts with directly now.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: Addressed. Thanks!

src/Adapters/EventSource.Netstandard13/EventSourceListeningRequest.cs
xiaomi7732: Good point. Will fix in next iteration.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: Thanks for the ask. Here's my consideration:

The assumptions:
1. This 'OnEventWritten' will be invoked huge amount of times per second.
2. DisabledSources's largely going to be empty.

Based on those, I want to reduce any context switch for the execution of the method whenever possible. Putting a pre-check, hopefully, it returns true, to avoid switching call-context to 'IsDroppingEvent' is the main reason.

---
Now that I think about it, having a pre-calculated list count shall further improve the performance and I'll update the code accordingly.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
SergeyKanzhelev: My understanding was the you'll include `Microsoft.ApplicaitonInsights` in the list of disabled sources by default as it causes the problem. If so - the list will never be empty and first loop will never result in a match. So I was suggesting to not run every matching rule twice - one for exact match and one for prefix match and do prefix match in the main loop.

xiaomi7732: @SergeyKanzhelev,
I don't think that will be the case if I understand the problem correctly.
My understanding is in most cases, there's no need to set DisabledSource at all unless there's event source starts to cause dead-loop.
@karolz-ms, could you please confirm?

karolz-ms: @xiaomi7732 correct, by default I would expect the list of disabled sources to be empty

SergeyKanzhelev: Wait. Isn't it the case that `Microsoft.ApplicationInsights` listener will cause problem all the time? And we do enable `Microsoft.ApplicationInsights` listener by default in Application Insights SDK

karolz-ms: No. In the report that originated this PR it was the Microsoft-ApplicationInsights-Data EventSource (a.k.a. RichPayloadEventSource) that triggered this, but typically either this event source is not used, or the buggy codepath in .NET EventSource/EventListener is not hit, and everything works.

So we are trying to be proactive here and give people a workaround *just in case they need it*.

That plus enabling or disabling sources based on prefix matching can cut the amount of configuration goo quite significantly, so it is a useful feature regardless of the .NET FX bug.

SergeyKanzhelev: ok. If config size a main concern - we can have a default list in code. If we know for sure it may happen and we know that this eventsource will never be listened from this telemetry module - it worth adding the default config.

This is how to initialize default list in code so it can be overridden from config or code later on: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/ecd7723b6de2a75530dc18741766825d6f5e42f5/Src/PerformanceCollector/Perf.Shared.NetFull/PerformanceCollectorModule.cs#L104-L111

xiaomi7732: @SergeyKanzhelev, thanks for the reference.

Here's my opinion:
1. I don't think the config size is the main concern.
2. We don't know which event source will be the one to cause the trouble until runtime - calling write event on dynamic EventSource basically will broadcast the event data to all the event source listeners enabled by the system (It doesn't matter if we registered to listen to that event source or not in this case) - the .net fx bug. Among those event sources, if there is one that will loop back to the code to trigger the same dynamic event, directly, or indirectly, a dead loop will be formed. Having the <DisabledSources /> will allow the user to break the loop in that scenario.
3. That said, I don't think we are failing by default. In most cases, the user shall not know the existence of <DisabledSources /> at all.
4. It is not a good idea to put anything into <DisabledSources /> by default because once there's anything in the collection, there will be performance degrade.

SergeyKanzhelev: My problem with this that if `Microsoft-ApplicationInsights-Data` enabled it guaranteed to be called as a result of ANY telemetry item reported by this telemetry module. Thus it is almost inevitably fall into this problem. Or I'm missing something?

 One potential solution will be to have some optimized check specifically for this event source.

karolz-ms: @SergeyKanzhelev the problem only manifested itself when a combination of EventFlow + ApplicationInsights + NLog was used, please refer to original issue https://github.com/Azure/diagnostics-eventflow/issues/140 . It does not happen when only EventSourceTelemetryModule with the rest of AI SDK is being used, otherwise EventSourceTelemetryModule would never work for anybody and the tests verifying its functionality would be failing all the time.

BUT

The problem was traced down to a bug in EventListener infrastructure https://github.com/dotnet/coreclr/issues/14434 I haven't debugged deeply enough into EventFlow + AI SDK + NLog combination to understand why this combination fails vs. EventFlow + AI SDK or EventSourceTelemetryModule + AI SDK work. But because I was able to reproduce the EventListener problem with a synthetic repro that has no EventFlow, AI SDK, or NLog,  we want to be proactive about giving EventSourceTelemetryModule users a workaround in case they hit it.

Since typical library combinations work, I think it is not necessary to disable the Microsoft-ApplicationInsights-Data EventSource by default.

Does this make sense now?

SergeyKanzhelev: Ok, I trust your judgment here.

SergeyKanzhelev:
@karolz-ms is there anything else you want fixed? Please sign off. It is looks OK for now.

src/Adapters.Tests/EventSource.Netstandard13.Tests/EventSourceTelemetryModuleTests.cs
```diff
@@ -85,6 +85,89 @@ public void ReportsSingleEvent()
 
         [TestMethod]
         [TestCategory("EventSourceListener")]
+        public void WildcardMatchEnablingEventSource()
```
karolz-ms: PrefixMatch

src/Adapters/EventSource.Netstandard13/EventSourceNamingMatchRuleBase.cs
```diff
@@ -0,0 +1,49 @@
...
+//     Copyright (c) Microsoft Corporation. All rights reserved.
+// </copyright>
+//-----------------------------------------------------------------------
+
+namespace Microsoft.ApplicationInsights.EventSourceListener
+{
+    /// <summary>
+    /// Abstraction of a match rule for event source. To be inherited by either an enabling rule or a disabling rule, etc.
+    /// </summary>
+    public abstract class EventSourceNamingMatchRuleBase
```
karolz-ms: Since there is is no "EventSourceNamingMatchRule" of any kind, let's name this EventSourceListeningRequestBase

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -65,6 +72,11 @@ public EventSourceTelemetryModule(OnEventWrittenHandler onEventWrittenHandler)
         public IList<EventSourceListeningRequest> Sources { get; private set; }
 
         /// <summary>
+        /// Gets the list of the EventSource disalbing rules in cases there's known event source that is causing issues.
```
karolz-ms: disabling requests

(typo + it is a collection of requests)

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -176,6 +192,21 @@ protected override void OnEventSourceCreated(EventSource eventSource)
         }
 
         /// <summary>
+        /// Invokes when change happens in DisabledSources list.
+        /// </summary>
+        private void DisabledEventSources_CollectionChanged(object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs e)
```
karolz-ms: My understanding is reacting to configuration changes in real time is not really expected from telemetry modules. What is expected is reacting to configuration changes when ITelemetryModule.Initialize() is called.

@SergeyKanzhelev do you have a preference?

@xiaomi7732 This scenario warrants a unit test (one that calls Initialize() twice and verifies that the behavior has changed according to changed configuration)

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -219,5 +250,50 @@ private void EnableAsNecessary(EventSource eventSource)
...
+        /// When the event comes from a EventSource that matches the rule of DisabledSource, it should be dropped, unless the event source name matches one of the 
+        /// name exactly in the enabling rule.
+        /// </summary>
+        /// <param name="eventData">The event data to test.</param>
+        private bool IsDroppingEvent(EventWrittenEventArgs eventData)
+        {
+            string eventSourceName = eventData?.EventSource?.Name;
+            if (string.IsNullOrEmpty(eventSourceName))
+            {
+                EventSourceListenerEventSource.Log.NullReference(nameof(EventSourceListener.EventSourceTelemetryModule), nameof(eventSourceName), $"{nameof(eventSourceName)} can't be null.");
```
karolz-ms: Just return false. I think it is possible to have an EventSource with just a GUID

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
karolz-ms: I agree. See https://github.com/Azure/diagnostics-eventflow/blob/master/src/Microsoft.Diagnostics.EventFlow.Inputs.EventSource/EventSourceInput.cs#L232 how I did it in EventFlow

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
```diff
@@ -219,5 +250,50 @@ private void EnableAsNecessary(EventSource eventSource)
...
+                return false;
+            }
+
+            // Do NOT drop the event when it matches one of the request by name exactly.
```
karolz-ms: Not sure why this check is necessary?

src/Adapters/EventSource.Shared/EventSource.Shared/Implementation/EventSourceListenerEventSource.cs
```diff
@@ -69,6 +70,12 @@ public void OnEventWrittenHandlerFailed(string moduleName, string details, strin
             this.WriteEvent(OnEventWrittenHandlerFailure, moduleName, details, applicationName ?? this.ApplicationName);
         }
 
+        [Event(NullReferenceEventId, Level =EventLevel.Error, Message ="Argument {1} is null in {0}.")]
```
karolz-ms: Please remove

xiaomi7732:
Great! I think we largely agree on the direction of moving this forward.
I'll work on the implementation details through the next few iterations.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: @karolz-ms,
The intention was to save the calculation when checking if there have been disable-listening requests.
Now I agree that it a better idea to have a snap of the decision than the current implementation. I'll update the code in the next iteration.

src/Adapters.Tests/EventSource.Netstandard13.Tests/EventSourceTelemetryModuleTests.cs
xiaomi7732: Will fix in the next iteration.

src/Adapters/EventSource.Netstandard13/EventSourceNamingMatchRuleBase.cs
xiaomi7732: Sure. Will be fixed in the next iteration.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: Corrected in the next iteration. Thanks!

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: I knew I forgot something and that is exactly the thing. Thanks for catching that. It will be fixed in the next iteration.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: Edge case would be the user has to disable **A**A, **A**B **A**C but needs to keep **A** work. I think that is agreed in #122. Let me know if you strongly disagree on this.

src/Adapters/EventSource.Shared/EventSource.Shared/Implementation/EventSourceListenerEventSource.cs
xiaomi7732: Done.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: As a matter of fact, I think it doesn't worth the complexity of the code. I'll get rid of it.

src/Adapters/EventSource.Netstandard13/EventSourceTelemetryModule.cs
xiaomi7732: Done.

xiaomi7732:
Code fix is done. Please feel free to let me know if you think more change is needed.

xiaomi7732:
@karolz-ms,
I think it is a good idea to employ a test case to cover the re-initialization scenario. I'll add it to the next iteration.

xiaomi7732:
@SergeyKanzhelev ,
I think all the concerns in the comments are addressed. Please sign off if everything looks good or let me know if there are further concerns. Thanks!

SergeyKanzhelev:
@cijothomas how can I merge this if builds wouldn't start for forks? Please help/advice

cijothomas:
merged.

cijothomas:
Can you update changelog.md to indicate this change ?

xiaomi7732:
@cijothomas 
Done!

cijothomas:
Thanks, merging.

PR #129, opened 2017-11-11
Flush 

SergeyKanzhelev:


src/Adapters.Tests/TraceListener.Tests.Shared/ApplicationInsightsTraceListenerTests.cs
```diff
@@ -367,6 +367,25 @@ public void TraceEventDoesNotStoreCallStackInTelemetryPropertiesBecauseLongValue
...
+            using (var listener = new ApplicationInsightsTraceListener(Guid.NewGuid().ToString()))
+            {
+                try
+                {
+                    listener.Flush();
+                    Assert.Fail();
+                }
+                catch (Exception ex)
+                {
+                    Assert.AreEqual("Flush called", ex.Message);
```
Dmitry-Matveev: This looks like rather unconventional way to test Flush(), something with `Assert.AreEqual(true, this.adapterHelper.Channel.Flushed)` will be more in line with the current approach.

PR #139, opened 2018-01-02
setup script for running binskim

MS-TimothyMothra:
BinSkim will fail on our test directories.
This script will copy necessary files to a new directory to be scanned

https://www.microsoft.com/en-us/SDL/adopt/tools.aspx

### DIFF
--- a/SDL/BinSkimSetup.ps1
+++ b/SDL/BinSkimSetup.ps1
@@ -1,0 +1,38 @@
+﻿[CmdletBinding()]
+Param(
+    [Parameter(Mandatory=$True)]
+    [string]$buildDirectory,
+    
+    [Parameter(Mandatory=$True)]
+    [string]$binSkimDirectory
+)
+
+#$buildDirectory = "C:\Repos\SDK\bin\Release\Src\"
+#$binSkimDirectory = "C:\Repos\SDK\binSkim\"
+
+Write-Host "`nPARAMETERS:";
+Write-Host "`tbuildDirectory:" $buildDirectory;
+Write-Host "`tbinSkimDirectory:" $binSkimDirectory;
+
+# don't need to clean folder on build server, but is needed for local dev
+if (Test-Path $binSkimDirectory) { Remove-Item $binSkimDirectory -Recurse; }
+
+# copy all
+Copy-Item -Path $buildDirectory -Filter "*.dll" -Destination $binSkimDirectory -Recurse;
+
+# delete test directories
+Get-ChildItem -Path $binSkimDirectory -Recurse -Directory | 
+ Where-Object {$_ -match "Test"} |
+ Remove-Item -Recurse;
+
+# summary for log output (file list and count)
+Write-Host "`nCOPIED FILES:";
+
+$count = 0;
+Get-ChildItem -Path $binSkimDirectory -Recurse -File | 
+    ForEach-Object { 
+        Write-Host "`t"$_.FullName; 
+        $count++;
+    } 
+
+Write-Host "`nTOTAL FILES:" $count;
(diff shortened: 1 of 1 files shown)


SDL/BinSkimSetup.ps1
```diff
@@ -0,0 +1,38 @@
+﻿[CmdletBinding()]
+Param(
+    [Parameter(Mandatory=$True)]
+    [string]$buildDirectory,
+    
+    [Parameter(Mandatory=$True)]
+    [string]$binSkimDirectory
+)
+
+#$buildDirectory = "C:\Repos\SDK\bin\Release\Src\"
```
d3r3kk: Suggestion: 
You can add default values to the parameters to make testing easier. Perhaps something like:

    [string]$buildDirectory=(Join-Path -Path (Split-Path -parent $PSCommandPath) -ChildPath "bin\Release\Src")

...might make testing locally a bit more handy.

SDL/BinSkimSetup.ps1
```diff
@@ -0,0 +1,38 @@
...
+#$binSkimDirectory = "C:\Repos\SDK\binSkim\"
+
+Write-Host "`nPARAMETERS:";
```
d3r3kk: Consider using 

    Write-Verbose "Parameters:"

here. Newline not really needed.

SDL/BinSkimSetup.ps1
```diff
@@ -0,0 +1,38 @@
...
+Write-Host "`tbuildDirectory:" $buildDirectory;
+Write-Host "`tbinSkimDirectory:" $binSkimDirectory;
+
+# don't need to clean folder on build server, but is needed for local dev
+if (Test-Path $binSkimDirectory) { Remove-Item $binSkimDirectory -Recurse; }
```
d3r3kk: Adding the following to your Test-Path will help out a bit in some cases:

    if (Test-Path -Path $binSkimDirectory -ErrorAction SilentlyContinue) { . . . }

SDL/BinSkimSetup.ps1
```diff
@@ -0,0 +1,38 @@
...
+
+# copy all
+Copy-Item -Path $buildDirectory -Filter "*.dll" -Destination $binSkimDirectory -Recurse;
+
+# delete test directories
+Get-ChildItem -Path $binSkimDirectory -Recurse -Directory | 
+ Where-Object {$_ -match "Test"} |
+ Remove-Item -Recurse;
```
d3r3kk: Do you not need a '-Force' switch on this Remove-Item?

SDL/BinSkimSetup.ps1
```diff
@@ -0,0 +1,38 @@
...
+
+# summary for log output (file list and count)
+Write-Host "`nCOPIED FILES:";
+
+$count = 0;
```
d3r3kk: Here's a shortcut:

    (Get-ChildItem -Path $binSkimDirectory -Recurse -File).Count

SDL/BinSkimSetup.ps1
MS-TimothyMothra: I understand the purpose of the Verbose distinction, but in this use case I don't think we want this. If a build fails, we would have to run it a second time with the verbose flag to get the details. 
I would rather have the logging always enabled.
Please let me know if you're considering something I'm not thinking of :)

SDL/BinSkimSetup.ps1
MS-TimothyMothra: THAT would make things easier!
But i need to go up three parents. Is there an easier way than this:

```
[string]$buildDirectory = (Join-Path -Path (Split-Path -parent (Split-Path -parent (Split-Path -parent $PSCommandPath))) -ChildPath "bin\Debug\Src") 
```

MS-TimothyMothra:
added script to copy bin DLLs to a new directory for BinSkim scanning tool

PR #141, opened 2018-01-10
sdl upgrade NLog

MS-TimothyMothra:
sdl has recommended but not required that we upgrade our version of NLog.

This is a trivial non-breaking change, no real reason not to do it.
This will not change the min-required version (4.3.8).

I ran all tests and they continue to pass.

### DIFF
--- a/src/NLogTarget/NLogTarget.csproj
+++ b/src/NLogTarget/NLogTarget.csproj
@@ -41,7 +41,7 @@
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.5.0-beta1" />
-    <PackageReference Include="NLog" Version="4.3.11" />
+    <PackageReference Include="NLog" Version="4.4.12" />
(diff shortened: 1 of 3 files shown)


src/NLogTarget/NLogTarget.csproj
```diff
@@ -41,7 +41,7 @@
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.5.0-beta1" />
-    <PackageReference Include="NLog" Version="4.3.11" />
+    <PackageReference Include="NLog" Version="4.4.12" />
```
SergeyKanzhelev: Make sure https://github.com/Microsoft/ApplicationInsights-dotnet-logging/blob/develop/ThirdPartyNotices.txt is up to date license-wise. Also mention this in CHANGELOG as this update make 4.3 of NLog not supported.

MS-TimothyMothra: Done and done.

PR #143, opened 2018-01-15
Fix NRE in DiagnosticSourceListener when property values are null

pharring:
The DiagnosticSourceListener tests project got removed with the last round of project shuffling (#135)
This puts it back.

Also fix a bug (NRE) when property values are null and add a test to cover that case.

### DIFF
--- a/Logging.sln
+++ b/Logging.sln
@@ -52,6 +52,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TraceListener", "src\TraceL
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "EventSource.Shared", "src\EventSource.Shared\EventSource.Shared\EventSource.Shared.shproj", "{A964DE6D-9750-4013-8BE2-79C2AFC056E5}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DiagnosticSourceListener.netcoreapp1.Tests", "test\DiagnosticSourceListener.netcoreapp10.Tests\DiagnosticSourceListener.netcoreapp1.Tests.csproj", "{625DABF6-C4AD-41A9-B2C0-04C9FEC2ADCF}"
+EndProject
 Global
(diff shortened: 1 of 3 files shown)


SergeyKanzhelev:
@pharring build failed with the error. Can you please fix?


3 error(s), 10 warning(s)
--
Build - 3 error(s), 10 warning(s)
src\DiagnosticSourceListener\DiagnosticSourceListenerSubscription.cs(61):src\DiagnosticSourceListener\DiagnosticSourceListenerSubscription.cs(61,0):   Error CA1305: Microsoft.Globalization : Because the behavior of   'Convert.ToString(object)' could vary based on the current user's locale   settings, replace this call in   'DiagnosticSourceListenerSubscription.OnNext(KeyValuePair<string,   object>)' with a call to 'Convert.ToString(object, IFormatProvider)'. If   the result of 'Convert.ToString(object, IFormatProvider)' will be displayed   to the user, specify System.Globalization.CultureInfo.CurrentCulture as the   'IFormatProvider' parameter. Otherwise, if the result will be stored and   accessed by software, such as when it is persisted to disk or to a database,   specify System.Globalization.CultureInfo.InvariantCulture.
src\DiagnosticSourceListener\DiagnosticSourceListenerSubscription.cs(61):src\DiagnosticSourceListener\DiagnosticSourceListenerSubscription.cs(61,0):   Error CA1305: Microsoft.Globalization : Because the behavior of   'Convert.ToString(object)' could vary based on the current user's locale   settings, replace this call in   'DiagnosticSourceListenerSubscription.OnNext(KeyValuePair<string,   object>)' with a call to 'Convert.ToString(object, IFormatProvider)'. If   the result of 'Convert.ToString(object, IFormatProvider)' will be displayed   to the user, specify 'CultureInfo.CurrentCulture' as the 'IFormatProvider'   parameter. Otherwise, if the result will be stored and accessed by software,   such as when it is persisted to disk or to a database, specify   'CultureInfo.InvariantCulture'.
Process   'msbuild.exe' exited with code '1'.

pharring:
@SergeyKanzhelev ugh! Sorry about that - I should have done a Release build.

PR #144, opened 2018-01-17
Use InvariantCulture to convert property values

pharring:
Fix Code Analysis breaks (CA1305) from #143


SergeyKanzhelev:
Thank you!

Issue #153, opened 2018-02-15
Use Authenticated User Id in Log4NetAppender only in Debug mode or not at all


Dmitry-Matveev:
We should remove automated Authenticated User ID collection from Log4NetAppender code [here](https://github.com/Microsoft/ApplicationInsights-dotnet-logging/blob/develop/src/Log4NetAppender/ApplicationInsightsAppender.cs#L145).

Having Authenticated User ID collected by default may present a privacy concern. However, it is acceptable to collect this for debug sessions to easily differentiate debug sessions between several users reporting to the same IKey. Acceptable does not mean necessary, so simple removal of User ID collection will do as well.

PR #158, opened 2018-02-27
remove autocollection of userId from Log4Net

MS-TimothyMothra:
Remove UserId from Log4Net. Update test and changelog.
#153

### DIFF
--- a/test/Log4NetAppender.Net45.Tests/ApplicationInsightsAppenderTests.cs
+++ b/test/Log4NetAppender.Net45.Tests/ApplicationInsightsAppenderTests.cs
@@ -162,20 +162,7 @@ namespace Microsoft.ApplicationInsights.Log4NetAppender.Tests
 
             Assert.IsNull(TelemetrySender.ValidateEndpointSend(telemetry));
         }
-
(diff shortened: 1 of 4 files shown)


test/Log4NetAppender.Net45.Tests/ApplicationInsightsAppenderTests.cs
```diff
@@ -162,20 +162,7 @@ public void TelemetryIsAcceptedByValidateEndpoint()
 
             Assert.IsNull(TelemetrySender.ValidateEndpointSend(telemetry));
         }
-
-        [TestMethod]
-        [TestCategory("Log4NetAppender")]
-        public void Log4NetSetsAuthenticatedUser()
```
Dmitry-Matveev: Would you mind changing this test into Log4NetDoesNotSetAuthenticatedUser() rather than remove it?
With this we will enforce this new requirement we've got.

MS-TimothyMothra: Makes sense. Done.

PR #163, opened 2018-03-12
prep 2.6.0 beta1

MS-TimothyMothra:


GlobalStaticVersion.props
```diff
@@ -6,14 +6,14 @@
       Update for every public release. 
     -->
     <SemanticVersionMajor>2</SemanticVersionMajor>
-    <SemanticVersionMinor>5</SemanticVersionMinor>
+    <SemanticVersionMinor>6</SemanticVersionMinor>
     <SemanticVersionPatch>0</SemanticVersionPatch>
-    <PreReleaseMilestone></PreReleaseMilestone>
+    <PreReleaseMilestone>beta1</PreReleaseMilestone>
```
Dmitry-Matveev: Would you like to make it beta2 as well to preserve consistency across all repos since it turned out we release beta2 everywhere anyway?

PR #176, opened 2018-04-18
NLog ApplicationInsightsTarget Flush with async delay

### LINKED ISSUE #177: NLog ApplicationInsightsTarget Flush should wait for completion
DefaultFlushDelay is 30 secs, and when flushed manually, then it happens in async-style, so one has to manually wait as specified in the documentation:

https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-custom-events-metrics#flushing-data

Since it is not currently possible to know when a Flush-operation actually completes, then one just have to make a quick guess of how long to wait, but also not too long.

See also https://stackoverflow.com/questions/49514679/no-log-entries-in-azure-application-insights-using-microsoft-applicationinsights

snakefoot:
But only when something to flush. Partially resolves #177 for NLog (Probably also an issue for all other logging destinations)

### DIFF
--- a/src/NLogTarget/ApplicationInsightsTarget.cs
+++ b/src/NLogTarget/ApplicationInsightsTarget.cs
@@ -28,6 +28,7 @@ namespace Microsoft.ApplicationInsights.NLogTarget
     public sealed class ApplicationInsightsTarget : TargetWithLayout
     {
         private TelemetryClient telemetryClient;
+        private DateTime lastLogEventTime;
 
         /// <summary>
         /// Initializers a new instance of ApplicationInsightsTarget type.
@@ -108,6 +109,8 @@ namespace Microsoft.ApplicationInsights.NLogTarget
                 throw new ArgumentNullException("logEvent");
             }
 
+            this.lastLogEventTime = DateTime.UtcNow;
+
             if (logEvent.Exception != null)
             {
                 this.SendException(logEvent);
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev:
Thank you for another contribution. We appreciate it.

For the housekeeping purposes - can you please create an issue with the problem description and then include changelog update in PR that would refer this issue.

snakefoot:
@SergeyKanzhelev Updated changelog for you

src/NLogTarget/ApplicationInsightsTarget.cs
```diff
@@ -127,7 +130,17 @@ protected override void FlushAsync(AsyncContinuation asyncContinuation)
...
-                asyncContinuation(null);
+                if (DateTime.UtcNow.AddSeconds(-30) > this.lastLogEventTime)
+                {
+                    // Nothing has been written, so nothing to wait for
+                    asyncContinuation(null);
+                }
+                else
+                {
+                    // Documentation says it is important to wait after flush, else nothing will happen
+                    // https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-custom-events-metrics#flushing-data
```
SergeyKanzhelev: please remove `en-US` from link so it will direct to localized version of the doc

src/NLogTarget/ApplicationInsightsTarget.cs
```diff
@@ -127,7 +130,17 @@ protected override void FlushAsync(AsyncContinuation asyncContinuation)
...
+                    System.Threading.Tasks.Task.Delay(500).ContinueWith((task) => asyncContinuation(null));
```
SergeyKanzhelev: Please use `TimeSpan.FromMilliseconds(500)` to make code a little bit more self-explanatory

SergeyKanzhelev:
I'll wait for other people to chime in and will merge tomorrow

src/NLogTarget/ApplicationInsightsTarget.cs
snakefoot: Fixed

src/NLogTarget/ApplicationInsightsTarget.cs
snakefoot: Fixed

snakefoot:
> It's a disappointment we need to work around this here, not make it a feature of base SDK

Yes it is a little weird not to have proper flush-logic in a logging framework.

Issue #178, opened 2018-04-26
Add 'CloudRoleName' property in 'ApplicationInsightsTarget.cs' to enable log filtering for multiple components


stpriyanka:
Need to add ``cloud_RoleName`` property to ``ApplicationInsightsTarget`` class. This property should enable filtering log by multiple components. 
It might be really small but very important issue in term of using same resource of ``Application insight`` for multiple components.

An overview of this solution is given:

````
 /// <summary>
 /// Gets or sets cloud_roleName
 /// </summary>
 public string Cloud_RoleName { get; set; }

 /// <summary>
 /// Gets or sets cloud_roleName
 /// </summary>
 public string Cloud_RoleInstance { get; set; } 
...........................

 protected override void InitializeTarget()
        {
            base.InitializeTarget();
            this.telemetryClient = new TelemetryClient();
          
            if (!string.IsNullOrEmpty(this.InstrumentationKey))
            {
                this.telemetryClient.Context.InstrumentationKey = this.InstrumentationKey;
            }

            if (!string.IsNullOrEmpty(this.Cloud_RoleName))
            {
                this.telemetryClient.Context.Cloud.RoleName = this.Cloud_RoleName;
            }

            if (!string.IsNullOrEmpty(this.Cloud_RoleInstance))
            {
                this.telemetryClient.Context.Cloud.RoleInstance = this.Cloud_RoleInstance;
            }

            this.telemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("nlog:");
        }
 
         ...........
````

PR #179, opened 2018-04-26
Added two properties on telemetryClient

naveednawaz:
#178 
In this way, we can add Cloud.RoleName and Cloud.RoleInstance

### DIFF
--- a/src/NLogTarget/ApplicationInsightsTarget.cs
+++ b/src/NLogTarget/ApplicationInsightsTarget.cs
@@ -44,6 +44,16 @@ namespace Microsoft.ApplicationInsights.NLogTarget
         public string InstrumentationKey { get; set; }
 
         /// <summary>
+        /// Gets or sets the Application Insights Cloud RoleName for your application. 
+        /// </summary>
+        public string CloudRoleName { get; set; }
+        
+         /// <summary>
+        /// Gets or sets the Application Insights Cloud RoleInstance for your application. 
+        /// </summary>
+        public string CloudRoleInstance { get; set; }
+        
+        /// <summary>
         /// Gets the logging controller we will be using.
         /// </summary>
         internal TelemetryClient TelemetryClient
@@ -94,7 +104,14 @@ namespace Microsoft.ApplicationInsights.NLogTarget
             {
                 this.telemetryClient.Context.InstrumentationKey = this.InstrumentationKey;
             }
-
+            if (!string.IsNullOrEmpty(this.CloudRoleName))
+            {
+                this.telemetryClient.Context.Cloud.RoleName = this.CloudRoleName;
+            }
+            if (!string.IsNullOrEmpty(this.CloudRoleInstance))
+            {
+                this.telemetryClient.Context.Cloud.RoleInstance = this.CloudRoleInstance;
+            }
             this.telemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("nlog:");
         }
 


pharring:
Consider what happens when those properties are modified after the telemetryClient Context has already been set in InitializeTarget. You could either have the setters throw or relay the value to telemetryClient.

SergeyKanzhelev:
@naveednawaz what is your scenario? Recommended way to set role name and role instance is to register corresponding telemetry initializer. Like:

``` csharp
TelemetryConfiguration.Active.TelemetryInitializers.Add(new DomainNameRoleInstanceTelemetryInitializer());
```

If you need to set custom values declaratively - potentially better solution will be to implement https://github.com/Microsoft/ApplicationInsights-dotnet/issues/630 which will allow to set role name and role instance for entire process. Or you can always implement simple custom initializer.

Drawback of proposed solution is that those properties will only be set on logs. Not on any other telemetry. So you'd loose ability to correlate different telemetry types.

naveednawaz:
@pharring  
I have added these properties exactly same like existing instrumentkey property. It's should work as instrumentkey

naveednawaz:
@SergeyKanzhelev 
I want  to use nlog target and add role name. Telemetry client is not accessible from applicationinsight target class.

SergeyKanzhelev:
@naveednawaz can you explain why setting role name globally like this wouldn't work? Like this:

``` csharp
TelemetryConfiguration.Active.TelemetryInitializers.Add(new DomainNameRoleInstanceTelemetryInitializer());
```

naveednawaz:
@SergeyKanzhelev
Can you tell me how can I add RoleName in this sample code by using 
TelemetryConfiguration.Active.TelemetryInitializers.Add(new DomainNameRoleInstanceTelemetryInitializer());

Console app (Install-Package Microsoft.ApplicationInsights.NLogTarget -Version 2.6.0-beta3)
var config = new LoggingConfiguration();
ApplicationInsightsTarget target = new ApplicationInsightsTarget();
// You need this only if you did not define InstrumentationKey in ApplicationInsights.config or want to use different instrumentation key
target.InstrumentationKey = "key";
LoggingRule rule = new LoggingRule("*", target);
config.LoggingRules.Add(rule);LogManager.Configuration = config;
Logger logger = LogManager.GetLogger("Example");
logger.Debug("trace log debug");

SergeyKanzhelev:
``` powershell
Install-Package Microsoft.ApplicationInsights.NLogTarget -Version 2.6.0-beta3
Install-Package Microsoft.ApplicationInsights.WindowsServer -Version 2.6.0-beta3
```

``` csharp
TelemetryConfiguration.Active.TelemetryInitializers.Add(new DomainNameRoleInstanceTelemetryInitializer());
var config = new LoggingConfiguration();
ApplicationInsightsTarget target = new ApplicationInsightsTarget();
// You need this only if you did not define InstrumentationKey in ApplicationInsights.config or want to use different instrumentation key
target.InstrumentationKey = "key";
LoggingRule rule = new LoggingRule("*", target);
config.LoggingRules.Add(rule);LogManager.Configuration = config;
Logger logger = LogManager.GetLogger("Example");
logger.Debug("trace log debug");
```


Alternatively you can define your own initializer:

``` powershell
Install-Package Microsoft.ApplicationInsights.NLogTarget -Version 2.6.0-beta3
```

``` csharp
public class CloudContextInitializer : ITelemetryInitializer
{
  public void Initialize(ITelemetry telemetry)
  {
    telemetry.Context.Cloud.RoleName = "role name";
    telemetry.Context.Cloud.RoleInstance = "role instance";
  }
}

TelemetryConfiguration.Active.TelemetryInitializers.Add(new CloudContextInitializer());
var config = new LoggingConfiguration();
ApplicationInsightsTarget target = new ApplicationInsightsTarget();
// You need this only if you did not define InstrumentationKey in ApplicationInsights.config or want to use different instrumentation key
target.InstrumentationKey = "key";
LoggingRule rule = new LoggingRule("*", target);
config.LoggingRules.Add(rule);LogManager.Configuration = config;
Logger logger = LogManager.GetLogger("Example");
logger.Debug("trace log debug");
```

SergeyKanzhelev:
@naveednawaz from my understanding of your scenario - proposed approach may be better that suggested changes. But you know your scenario the best - thus my questions to better understand what's missing.

naveednawaz:
@SergeyKanzhelev 
Thanks for your reply. I tried exactly you mentioned but it's not working. Initialize didn't hit. I'm using .net core 2 console application. I really appreciate if you can try and suggest  working solution.

stpriyanka:
Well, that logic works for me only after adding ```min/max LogLevel``` or both to trigger this rule.
 ```  var rule = new LoggingRule("*", NLog.LogLevel.Trace, LogLevel.Fatal, target); ```

SergeyKanzhelev:
I'm closing this PR. If initializers are not being called - please file an issue for this to investigate

PR #181, opened 2018-04-28
Prevent double telemetry if DiagnosticSourceTelemetryModule is initialized twice

pharring:
Prevent double telemetry if DiagnosticSourceTelemetryModule is initialied twice.

Fixes https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/653

### DIFF
--- a/src/DiagnosticSourceListener/DiagnosticSourceTelemetryModule.cs
+++ b/src/DiagnosticSourceListener/DiagnosticSourceTelemetryModule.cs
@@ -9,6 +9,7 @@ namespace Microsoft.ApplicationInsights.DiagnosticSourceListener
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Threading;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
     using Microsoft.ApplicationInsights.Implementation;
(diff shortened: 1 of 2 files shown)


src/DiagnosticSourceListener/DiagnosticSourceTelemetryModule.cs
```diff
@@ -38,9 +38,16 @@ public void Initialize(TelemetryConfiguration configuration)
...
-            this.client = new TelemetryClient(configuration);
-            this.client.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("dsl:");
-            this.allDiagnosticListenersSubscription = DiagnosticListener.AllListeners.Subscribe(this);
+            var telemetryClient = new TelemetryClient(configuration);
+            telemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("dsl:");
+
+            this.client = telemetryClient;
+
+            // Protect against multiple subscriptions if Initialize is called twice
+            if (this.allDiagnosticListenersSubscription == null)
```
SergeyKanzhelev: should it be locked like here: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/8645f55e16284a8468c40bc34c2ca7091e92b547/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs#L121-L126

pharring: OK. Uses CAS now.

PR #162, opened 2018-03-08
add fxcop analyzer

MS-TimothyMothra:
Add the FxCop Analyzer to Non-Test projects.

This analyzer identifies **SEVERAL** compiler errors. 
If you're bored one afternoon, feel free to pull this branch and fix a couple! :)

dreamwork

PR #182, opened 2018-04-29
Fix FxCop violations

pharring:
CA1307 Use StringComparison.Ordinal for StartsWith
CA1063 Seal classes with trivial Dispose methods
CA1305 Use InvariantCulture for ToString and string.Format
CA1309 Use ordinal comparison
CA1507 Use nameof
CA1801 Review unused parameters
CA1823 Remove unused fields
CA1825 Avoid zero length array allocations
CA2007 Call ConfigureAwait


pharring:
See #162 
This isn't all of them, but it's a good chunk.

PR #152, opened 2018-02-15
Add NLog GlobalDiagnosticContext properties.

jeff-zohrab:
Nlog has a Global Diagnostics Context dictionary for app-wide
instance values (ref https://github.com/NLog/NLog/wiki/Gdc-layout-renderer).
Include these in the properties sent to the NLog target.

PR #183, opened 2018-04-29
GlobalDiagnosticContext must be explictly added using ContextProperties

snakefoot:
This is a refactoring of #152 by @jeff-zohrab

It is a bad idea to include all GDC-items, as they are often used a global variables across all NLog-targets. This means all kind of random junk will suddenly be sent to the ApplicationInsightsTarget.

Instead this PR allows one to add the wanted global items explicitly using the new ContextProperties:

```xml
<nlog>
   <extensions>
      <add assembly="Microsoft.ApplicationInsights.NLogTarget" />
   </extensions>
   <targets>
      <target type="ApplicationInsightsTarget" name="aiTarget" />
           <contextProperty name="appid" layout="${gdc:item=appid}" />
           <contextProperty name="threadid" layout="${threadid}" />
      </target>
   </targets>
    <rules>
      <logger name="*" minlevel="Trace" writeTo="aiTarget" />
    </rules>
</nlog>
```

One can also add other random stuff like threadid, trace-correlationid, etc.


snakefoot:
One might consider removing this code:

https://github.com/Microsoft/ApplicationInsights-dotnet-logging/blob/388ff8e884711ffdc6a7ef8cab16a0b136b6c2ac/src/NLogTarget/ApplicationInsightsTarget.cs#L75-L79

As it can be replaced by:
```xml
      <target type="ApplicationInsightsTarget" name="aiTarget" />
           <contextProperty name="ClassName" layout="${callsite:className=true:fileName=false:methodName=false}" />
           <contextProperty name="FileName" layout="${callsite:className=false:fileName=true:methodName=false}" />
           <contextProperty name="MethodName" layout="${callsite:className=false:fileName=false:methodName=true}" />
           <contextProperty name="LineNumber" layout="${callsite-linenumber}" />
      </target>
```
And it will actually ensure that NLog will capture a useful UserStackFrame.

SergeyKanzhelev:
@snakefoot thanks for another good contribution!

@jeff-zohrab are you around to provide feedback?

jeff-zohrab:
Hi @SergeyKanzhelev , I'll look through the code.  Thanks @snakefoot, this improves my initial submission.

test/NLogTarget.Net45.Tests/NLogTargetTests.cs
```diff
@@ -222,7 +222,9 @@ public void TraceHasCustomProperties()
         [TestCategory("NLogTarget")]
         public void GlobalDiagnosticContextPropertiesAreAddedToProperties()
```
jeff-zohrab: @snakefoot - consider renaming test: `GlobalDiagnosticContextPropertiesAreAddedToPropertiesIfContextIsSet`.  That's a long method name, but it clarifies the change you made.

This also suggests a test: `GlobalDiagnosticContextPropertiesAreNotAddedToPropertiesIfContextNotSet`. Which is an even longer method name.

jeff-zohrab:
@snakefoot - this is an improvement on my submission, thanks.  I suggested one revision to the test name only.  Looks good!

test/NLogTarget.Net45.Tests/NLogTargetTests.cs
snakefoot: Think it is a little strange to have unit-test that verifies functionality is not there. Have no opinion about the unit test names. Just happy that they include `GlobalDiagnosticContext` in the name.

jeff-zohrab: OK, will leave the decision to @SergeyKanzhelev, it's fine either way.

PR #203, opened 2018-06-27
NLog ApplicationInsightsTarget - Add support for using GDC for InstrumentationKey

### LINKED ISSUE #201: [NLOG] Add layout support for the ApplicationInsightsTarget target
Package: Microsoft.ApplicationInsights.NLogTarget, v2.6.4

Scenario: we want to configure the target via xml file and pass instrumentation key obtained from azure key vault.

NLog supports dynamic values via property layouts. Usually, most targets do support things like 
```
// Nlog.config
		<target type="ApplicationInsightsTarget" name="aiTarget" instrumentationKey="${gdc:item=AppInsightsInstrumentationKey}" />

// code
NLog.GlobalDiagnosticsContext.Set(
		"AppInsightsInstrumentationKey",
		instrumentationKeyFromKeyVault);
```

ApplicationInsightsTarget target passes its `instrumentationKey` property as is so the AppInsightsInstrumentationKey value is not expanded.

Please, use Render() for the `InstrumentationKey` property.

snakefoot:
Resolves #201

Normally NLog properties are just added like this:

`public NLog.Layouts.Layout InstrumentationKey { get; set; }`

But because it would be a breaking change, then this PR is a little convoluted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog 
 
-
+### Version 2.6.5
+- [NLog can perform Layout of InstrumentationKey](https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/203)
 
 ### Version 2.6.4
 - [Log4Net new supports NetStandard 1.3!](https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/167)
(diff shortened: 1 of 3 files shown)


snakefoot:
Maybe you can merge #194 first?



Sent from my Samsung device


-------- Original message --------

SergeyKanzhelev:
@snakefoot done

snakefoot:
@SergeyKanzhelev Updated ChangeLog.md

PR #217, opened 2018-09-13
disable fxcop "Identifiers should not match keywords"

MS-TimothyMothra:


ApplicationInsightsSDKRules.ruleset
```diff
@@ -25,6 +25,8 @@
     <Rule Id="CA1403" Action="Warning" />
     <Rule Id="CA1405" Action="Warning" />
     <Rule Id="CA1410" Action="Warning" />
+    <!-- Suppressing Message "dentifiers should not match keywords" because class TplActivities is part of our public api and can't be changed -->
```
Dmitry-Matveev: minor: identifiers

PR #221, opened 2018-09-19
[WIP] moved ILogger as is

SergeyKanzhelev:
This PR is to collect the feedback

CC: @davidfowl

### DIFF
--- a/src/ILogger/ApplicationInsightsLoggerFactoryExtensions.cs
+++ b/src/ILogger/ApplicationInsightsLoggerFactoryExtensions.cs
@@ -1,0 +1,87 @@
+﻿namespace Microsoft.Extensions.Logging
+{
+    using System;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.ILogger;
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Options;
+
+    /// <summary>
+    /// Extension methods for <see cref="ILoggerFactory"/> that allow adding Application Insights logger.
+    /// </summary>
+    public static class ApplicationInsightsLoggerFactoryExtensions
+    {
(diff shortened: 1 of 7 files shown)


src/ILogger/ApplicationInsightsLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+    using System;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.ILogger;
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Options;
+
+    /// <summary>
+    /// Extension methods for <see cref="ILoggerFactory"/> that allow adding Application Insights logger.
+    /// </summary>
+    public static class ApplicationInsightsLoggerFactoryExtensions
```
davidfowl: I'd like to remove these extension methods completely and use the ILoggerBuilder pattern. It better fits what AI because the telemetry client needs to be added.

davidfowl:
@pakrym Can you a review here? I have some changes locally that re-baselines on .NET Standard 2.0 and the 2.0 version of logging. Where I'm currently stuck is trying to figure out how to implement AddApplicationInsights on ILoggerBuilder. We need a bunch of services here and I was messing with this:

```C#
public static ILoggingBuilder AddApplicationInsights(this ILoggingBuilder builder)
{
    builder.Services.TryAddSingleton<TelemetryClient>();
    builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<ILoggerProvider, ApplicationInsightsLoggerProvider>());
    return builder;
}
```

This needs to compose with the method in the ASP.NET Core SDK so that things work if you call AddApplicationInsights directly or UseApplicationInsights on the IWebHostBuilder.

PR #239, opened 2018-12-07
ILogger implementation for ApplicationInsights

RamjotSingh:
Original discussion at -  https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/749

### C# Console program sample

Packages installed 

```xml
<ItemGroup>
    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.1.0" />
    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.0" />
    <PackageReference Include="Microsoft.Extensions.Logging.ApplicationInsights" Version="2.8.1-build39946" />
    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="2.1.0" />
  </ItemGroup>
```

```cs
class Program
    {
        static void Main(string[] args)
        {
            // Create DI container.
            IServiceCollection services = new ServiceCollection();
            
            // Add the logging pipelines to use. We are using Console and AI and configuring them both.
            services.AddLogging(loggingBuilder =>
            {
                loggingBuilder.AddApplicationInsights("--YourAIKeyHere--");

                loggingBuilder.AddConsole();
            });

            // Build ServiceProvider.
            IServiceProvider serviceProvider = services.BuildServiceProvider();

            ILogger<Program> logger = serviceProvider.GetRequiredService<ILogger<Program>>();

            // Begin a new scope. This is optional. Epecially in case of AspNetCore request info is already
            // present in scope.
            using (logger.BeginScope(new Dictionary<string, object> { { "Method", nameof(Main) } }))
            {
                logger.LogInformation("Logger is working");
            }
        }
    }
```

### DIFF
--- a/Logging.sln
+++ b/Logging.sln
@@ -63,6 +63,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NLogTarget.NetCoreApp10.Tes
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Log4NetAppender.NetCoreApp10.Tests", "test\Log4NetAppender.NetCoreApp10.Tests\Log4NetAppender.NetCoreApp10.Tests.csproj", "{F74826DB-53B1-4588-BD02-4DD25DCBF292}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ILogger", "src\ILogger\ILogger.csproj", "{7D942802-E85E-4C3C-B97E-67A3867B7CBA}"
+EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		test\CommonTestShared\CommonTestShared.projitems*{3b9ab7fa-562d-4e4e-86e3-3348426bc0d9}*SharedItemsImports = 13
@@ -225,6 +227,14 @@ Global
 		{F74826DB-53B1-4588-BD02-4DD25DCBF292}.Release|Any CPU.Build.0 = Release|Any CPU
 		{F74826DB-53B1-4588-BD02-4DD25DCBF292}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
 		{F74826DB-53B1-4588-BD02-4DD25DCBF292}.Release|Mixed Platforms.Build.0 = Release|Any CPU
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Release|Any CPU.Build.0 = Release|Any CPU
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Release|Mixed Platforms.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -251,6 +261,7 @@ Global
 		{11E4A92F-154E-450B-8508-437C28744BC2} = {2FCC45B3-D820-405D-87FA-467C96465BB1}
 		{B24CB3C6-A3A0-4190-97E4-E847C7B1A691} = {2FCC45B3-D820-405D-87FA-467C96465BB1}
 		{F74826DB-53B1-4588-BD02-4DD25DCBF292} = {2FCC45B3-D820-405D-87FA-467C96465BB1}
+		{7D942802-E85E-4C3C-B97E-67A3867B7CBA} = {EE933574-C82B-4E59-A0D1-05328197B937}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {AC8888B1-0E98-49D7-BE15-EB97A6261C0D}
--- a/Logging.sln
+++ b/Logging.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2036
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28315.86
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{EE933574-C82B-4E59-A0D1-05328197B937}"
 EndProject
@@ -63,7 +63,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NLogTarget.NetCoreApp10.Tes
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Log4NetAppender.NetCoreApp10.Tests", "test\Log4NetAppender.NetCoreApp10.Tests\Log4NetAppender.NetCoreApp10.Tests.csproj", "{F74826DB-53B1-4588-BD02-4DD25DCBF292}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ILogger", "src\ILogger\ILogger.csproj", "{7D942802-E85E-4C3C-B97E-67A3867B7CBA}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ILogger", "src\ILogger\ILogger.csproj", "{7D942802-E85E-4C3C-B97E-67A3867B7CBA}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ILogger.NetStandard.Tests", "test\ILogger.NetStandard.Tests\ILogger.NetStandard.Tests.csproj", "{4C1B862F-8F19-4D0F-834D-694DDE0438AE}"
 EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
@@ -235,6 +237,14 @@ Global
 		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Release|Any CPU.Build.0 = Release|Any CPU
 		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
 		{7D942802-E85E-4C3C-B97E-67A3867B7CBA}.Release|Mixed Platforms.Build.0 = Release|Any CPU
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE}.Release|Any CPU.Build.0 = Release|Any CPU
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE}.Release|Mixed Platforms.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -262,6 +272,7 @@ Global
 		{B24CB3C6-A3A0-4190-97E4-E847C7B1A691} = {2FCC45B3-D820-405D-87FA-467C96465BB1}
 		{F74826DB-53B1-4588-BD02-4DD25DCBF292} = {2FCC45B3-D820-405D-87FA-467C96465BB1}
 		{7D942802-E85E-4C3C-B97E-67A3867B7CBA} = {EE933574-C82B-4E59-A0D1-05328197B937}
+		{4C1B862F-8F19-4D0F-834D-694DDE0438AE} = {2FCC45B3-D820-405D-87FA-467C96465BB1}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {AC8888B1-0E98-49D7-BE15-EB97A6261C0D}
--- a/src/ILogger/ApplicationInsightsLogger.cs
+++ b/src/ILogger/ApplicationInsightsLogger.cs
@@ -12,7 +12,6 @@ namespace Microsoft.Extensions.Logging.ApplicationInsights
     using System.Globalization;
     using System.Text;
     using Microsoft.ApplicationInsights;
-    using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
     using Microsoft.ApplicationInsights.Implementation;
@@ -87,7 +86,7 @@ namespace Microsoft.Extensions.Logging.ApplicationInsights
                 if (exception == null || !this.applicationInsightsLoggerOptions.TrackExceptionsAsExceptionTelemetry)
                 {
                     TraceTelemetry traceTelemetry = new TraceTelemetry(
-                        formatter(state, exception), 
+                        formatter(state, exception),
                         ApplicationInsightsLogger.GetSeverityLevel(logLevel));
                     this.PopulateTelemetry(traceTelemetry, state, eventId);
                     this.telemetryClient.TrackTrace(traceTelemetry);
@@ -99,7 +98,7 @@ namespace Microsoft.Extensions.Logging.ApplicationInsights
                         Message = formatter(state, exception),
                         SeverityLevel = ApplicationInsightsLogger.GetSeverityLevel(logLevel),
                     };
-                    exceptionTelemetry.Context.Properties["Exception"] = exception.ToString();
+
                     this.PopulateTelemetry(exceptionTelemetry, state, eventId);
                     this.telemetryClient.TrackException(exceptionTelemetry);
                 }
@@ -151,19 +150,19 @@ namespace Microsoft.Extensions.Logging.ApplicationInsights
         /// Populates the state, scope and event information for the logging event.
         /// </summary>
         /// <typeparam name="TState">State information for the current event.</typeparam>
-        /// <param name="telemetry">Telemetry item.</param>
+        /// <param name="telemetryItem">Telemetry item.</param>
         /// <param name="state">Event state information.</param>
         /// <param name="eventId">Event Id information.</param>
-        private void PopulateTelemetry<TState>(ITelemetry telemetry, TState state, EventId eventId)
+        private void PopulateTelemetry<TState>(ISupportProperties telemetryItem, TState state, EventId eventId)
         {
-            IDictionary<string, string> dict = telemetry.Context.Properties;
+            IDictionary<string, string> dict = telemetryItem.Properties;
             dict["CategoryName"] = this.categoryName;
 
-            if (this.applicationInsightsLoggerOptions.IncludeEventId)
+            if (!this.applicationInsightsLoggerOptions.ExcludeEventId)
             {
                 if (eventId.Id != 0)
                 {
-                    dict["EventId"] = eventId.Id.ToString(System.Globalization.CultureInfo.InvariantCulture);
+                    dict["EventId"] = eventId.Id.ToString(CultureInfo.InvariantCulture);
                 }
 
                 if (!string.IsNullOrEmpty(eventId.Name))
@@ -172,41 +171,44 @@ namespace Microsoft.Extensions.Logging.ApplicationInsights
                 }
             }
 
-            if (state is IReadOnlyList<KeyValuePair<string, object>> stateDictionary)
+            if (!this.applicationInsightsLoggerOptions.ExcludeScopeData)
             {
-                foreach (KeyValuePair<string, object> item in stateDictionary)
+                if (state is IReadOnlyList<KeyValuePair<string, object>> stateDictionary)
                 {
-                    dict[item.Key] = Convert.ToString(item.Value, CultureInfo.InvariantCulture);
-                }
-            }
-
-            if (this.ExternalScopeProvider != null)
-            {
-                StringBuilder stringBuilder = new StringBuilder();
-                this.ExternalScopeProvider.ForEachScope(
-                    (activeScope, builder) =>
+                    foreach (KeyValuePair<string, object> item in stateDictionary)
                     {
-                        // Ideally we expect that the scope to implement IReadOnlyList<KeyValuePair<string, object>>.
-                        // But this is not guaranteed as user can call BeginScope and pass anything. Hence
-                        // we try to resolve the scope as Dictionary and if we fail, we just serialize the object and add it.
+                        dict[item.Key] = Convert.ToString(item.Value, CultureInfo.InvariantCulture);
+                    }
+                }
 
-                        if (activeScope is IReadOnlyList<KeyValuePair<string, object>> activeScopeDictionary)
-                        {
-                            foreach (KeyValuePair<string, object> item in activeScopeDictionary)
-                            {
-                                dict[item.Key] = Convert.ToString(item.Value, CultureInfo.InvariantCulture);
-                            }
-                        }
-                        else
-                        {
-                            stringBuilder.Append(" => ").Append(activeScope);
-                        }
-                    },
-                    stringBuilder);
-
-                if (stringBuilder.Length > 0)
+                if (this.ExternalScopeProvider != null)
                 {
-                    dict["Scope"] = stringBuilder.ToString();
+                    StringBuilder stringBuilder = new StringBuilder();
+                    this.ExternalScopeProvider.ForEachScope(
+                        (activeScope, builder) =>
+                        {
+                            // Ideally we expect that the scope to implement IReadOnlyList<KeyValuePair<string, object>>.
+                            // But this is not guaranteed as user can call BeginScope and pass anything. Hence
+                            // we try to resolve the scope as Dictionary and if we fail, we just serialize the object and add it.
+
+                            if (activeScope is IReadOnlyList<KeyValuePair<string, object>> activeScopeDictionary)
+                            {
+                                foreach (KeyValuePair<string, object> item in activeScopeDictionary)
+                                {
+                                    dict[item.Key] = Convert.ToString(item.Value, CultureInfo.InvariantCulture);
+                                }
+                            }
+                            else
+                            {
+                                builder.Append(" => ").Append(activeScope);
+                            }
+                        },
+                        stringBuilder);
+
+                    if (stringBuilder.Length > 0)
+                    {
+                        dict["Scope"] = stringBuilder.ToString();
+                    }
                 }
             }
         }
--- a/src/ILogger/ApplicationInsightsLogger.cs
+++ b/src/ILogger/ApplicationInsightsLogger.cs
@@ -1,0 +1,203 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="ApplicationInsightsLogger.cs" company="Microsoft">
+// Copyright (c) Microsoft Corporation. 
+// All rights reserved.  2013
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Microsoft.Extensions.Logging.ApplicationInsights
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Text;
+    using Microsoft.ApplicationInsights;
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Implementation;
+
+    /// <summary>
+    /// Application insights logger implementation for <see cref="ILogger"/>.
+    /// </summary>
+    /// <seealso cref="Microsoft.Extensions.Logging.ILogger" />
+    public class ApplicationInsightsLogger : ILogger, IDisposable
+    {
+        private readonly string categoryName;
+        private readonly TelemetryClient telemetryClient;
+        private readonly ApplicationInsightsLoggerOptions applicationInsightsLoggerOptions;
+
+        /// <summary>
+        /// Creates a new instance of <see cref="ApplicationInsightsLogger"/>
+        /// </summary>
+        public ApplicationInsightsLogger(
+            string categoryName,
+            TelemetryClient telemetryClient,
+            ApplicationInsightsLoggerOptions applicationInsightsLoggerOptions)
+        {
+            this.categoryName = categoryName;
+            this.telemetryClient = telemetryClient;
+            this.telemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("il:");
+            this.applicationInsightsLoggerOptions = applicationInsightsLoggerOptions ?? throw new ArgumentNullException(nameof(applicationInsightsLoggerOptions));
+        }
+
+        /// <summary>
+        /// Gets or sets the external scope provider.
+        /// </summary>
+        internal IExternalScopeProvider ExternalScopeProvider { get; set; }
+
+        /// <summary>
+        /// Begins a logical operation scope.
+        /// </summary>
+        /// <typeparam name="TState">Current state.</typeparam>
+        /// <param name="state">The identifier for the scope.</param>
+        /// <returns>
+        /// An IDisposable that ends the logical operation scope on dispose.
+        /// </returns>
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            return this.ExternalScopeProvider != null ? this.ExternalScopeProvider.Push(state) : NullScope.Instance;
+        }
+
+        /// <summary>
+        /// Checks if the given <paramref name="logLevel" /> is enabled.
+        /// </summary>
+        /// <param name="logLevel">level to be checked.</param>
+        /// <returns>
+        ///   <c>true</c> if enabled.
+        /// </returns>
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return this.telemetryClient.IsEnabled();
+        }
+
+        /// <summary>
+        /// Writes a log entry.
+        /// </summary>
+        /// <typeparam name="TState">State being passed along.</typeparam>
+        /// <param name="logLevel">Entry will be written on this level.</param>
+        /// <param name="eventId">Id of the event.</param>
+        /// <param name="state">The entry to be written. Can be also an object.</param>
+        /// <param name="exception">The exception related to this entry.</param>
+        /// <param name="formatter">Function to create a <c>string</c> message of the <paramref name="state" /> and <paramref name="exception" />.</param>
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (this.IsEnabled(logLevel))
+            {
+                if (exception == null || this.applicationInsightsLoggerOptions?.TrackExceptionsAsExceptionTelemetry == false)
+                {
+                    TraceTelemetry traceTelemetry = new TraceTelemetry(formatter(state, exception), this.GetSeverityLevel(logLevel));
+                    this.PopulateTelemetry(traceTelemetry, state, eventId);
+                    this.telemetryClient.TrackTrace(traceTelemetry);
+                }
+                else
+                {
+                    ExceptionTelemetry exceptionTelemetry = new ExceptionTelemetry(exception)
+                    {
+                        Message = formatter(state, exception),
+                        SeverityLevel = this.GetSeverityLevel(logLevel),
+                    };
+                    exceptionTelemetry.Context.Properties["Exception"] = exception.ToString();
+                    this.PopulateTelemetry(exceptionTelemetry, state, eventId);
+                    this.telemetryClient.TrackException(exceptionTelemetry);
+                }
+            }
+        }
+
+        /// <summary>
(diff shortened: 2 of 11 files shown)


RamjotSingh:
Trying to collect all the issues which are logged across various repos. Please feel free to reference other issues here.

https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/491
https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/749
https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/783

lmolkova:
@RamjotSingh we have an ongoing conversation with Asp.Net Core team on the high-level design of this feature . There are many things to consider: scopes support and meaning in AppInsights, sampling, common ids.

We also have ILogger implementation in AspNEtCore repo: https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/ed794ae3a0150715d76432c7417aa93c334126af/src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLogger.cs

So could we hold this on until we finalize the design?

cc @SergeyKanzhelev

RamjotSingh:
@lmolkova Oh ok that's cool. I thought that this was sitting on cold plate so I ended up picking it up. The ILogger implementation I have done is actually based on what @SergeyKanzhelev wrote in his private branch (which was in turn based on AI-ASPNetCore repo). If you feel this is getting picked up I can close the PR for now. 

Let me know what you think.

cijothomas:
@RamjotSingh Thanks for your effort here - we definitely want to take it. We may change implementation later based on discussions with asp.net core team, but moving this to logging is the right thing to do now!
Will review this shortly

cijothomas:
Once we merge this and ship a new nuget package from this repo (Microsoft.ApplicationInsights.ILoggerProvider or Microsoft.Extensions.Logging.ApplicationInsights), the extension methods on application insights asp.net core sdk can be marked obsolete to be removed in the next major version bump. It'll still contain the duplicate code for ApplicationInsightsLogger etc.

User who want to use ILogger with ApplicationInsights should use this new package, and new extension methods.

src/ILogger/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,203 @@
...
+                    this.telemetryClient.TrackTrace(traceTelemetry);
+                }
+                else
+                {
+                    ExceptionTelemetry exceptionTelemetry = new ExceptionTelemetry(exception)
+                    {
+                        Message = formatter(state, exception),
+                        SeverityLevel = this.GetSeverityLevel(logLevel),
+                    };
+                    exceptionTelemetry.Context.Properties["Exception"] = exception.ToString();
```
pharring: What is the point of this? Isn't the exception object itself enough?

src/ILogger/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,203 @@
...
+                    stringBuilder);
+
+                if (stringBuilder.Length > 0)
+                {
+                    dict["Scope"] = stringBuilder.ToString();
+                }
+            }
+        }
+
+        private SeverityLevel GetSeverityLevel(LogLevel logLevel)
```
pharring: static

src/ILogger/ApplicationInsightsLogger.cs
RamjotSingh: I picked this part of code from @SergeyKanzhelev 's branch. Who in turn IIRC picked from the existing ILogger implementation for ASP.NET Core. Changing this implementation might change the behavior of people using ASP.NET Core once they upgrade. If we are fine with it. I can look into changing this.

src/ILogger/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,203 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="ApplicationInsightsLogger.cs" company="Microsoft">
+// Copyright (c) Microsoft Corporation. 
+// All rights reserved.  2013
```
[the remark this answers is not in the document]
RamjotSingh: Been wondering myself. Checked the existing files, they said 2013 as well, so left it as is. Also I have rarely seen other AI repos to use file header so not sure what to make of it.

src/ILogger/ApplicationInsightsLogger.cs
funkycoding: See https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/574 for some background info on above code

RamjotSingh: I think @pharring 's point was that why are we double logging the exception object. Thing is ExceptionTelemetry is parsing the stack trace and then pulling information from it and putting it in parsedStack. Adding this property gives you back to full Exception few that you would normally get when IDE (VS) shows it. There are advantages of adding this and disadvantages.

Advantage is you get the full stack that you would have otherwise have seen. However this information is already present in the parsed stack. So we are not really gaining anything very special. Also this is adding a lot of bits over the wire and thus will contribute to user's quota.

Based on the existing non ILogger implementation [here](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/2e5aa6c9cf787123990cf56a2505d7137f1704e5/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs#L474). I would recommend we remove this. We can re-add this if people ask for it. This will obviously cause the new logger to have different behavior than the existing ILogger. So I will let AI team decide on it.

RamjotSingh: Just for reference here is what the payload over the wire looks like if we include **Exception** property


```json
{
    "name": "Microsoft.ApplicationInsights.Dev.00000000000000000000000000000000.Exception",
    "time": "2018-12-17T11:27:59.7453991Z",
    "iKey": "00000000-0000-0000-0000-000000000000",
    "tags": {
        "ai.application.ver": "1.0.0.0",
        "ai.cloud.roleInstance": "---",
        "ai.operation.id": "b841d3050a026441b3c8ad9e6e8b7526",
        "ai.operation.parentId": "|b841d3050a026441b3c8ad9e6e8b7526.d7fff5e7_",
        "ai.operation.name": "GET Values/Get",
        "ai.location.ip": "::1",
        "ai.internal.sdkVersion": "il:2.8.1-36930",
        "ai.internal.nodeName": "---"
    },
    "data": {
        "baseType": "ExceptionData",
        "baseData": {
            "ver": 2,
            "exceptions": [{
                "id": 29403036,
                "outerId": 0,
                "typeName": "System.ArgumentNullException",
                "message": "An unhandled exception has occurred while executing the request.",
                "hasFullStack": true,
                "parsedStack": [{
                    "level": 0,
                    "method": "ApplicationInsightsModernLogger.Controllers.ValuesController.Get",
                    "assembly": "ApplicationInsightsModernLogger, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null",
                    "fileName": "D:\\Code\\VSProjects\\ApplicationInsightsModernLogger\\ApplicationInsightsModernLogger\\Controllers\\ValuesController.cs",
                    "line": 27
                }, {
                    "level": 1,
                    "method": "lambda_method",
                    "assembly": "Anonymously Hosted DynamicMethods Assembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null",
                    "line": 0
                }, ...... REST OF THE PARSED STACK TRACE
            }],
            "severityLevel": "Error",
            "properties": {
                "CategoryName": "Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware",
                "RequestId": "0HLJ42AH8F859:00000001",
                "AspNetCoreEnvironment": "Development",
                "RequestPath": "/api/values",
                "DeveloperMode": "true",
                "{OriginalFormat}": "An unhandled exception has occurred while executing the request.",
                "Exception": "System.ArgumentNullException: Value cannot be null.\r\nParameter name: testParam\r\n   at ApplicationInsightsModernLogger.Controllers.ValuesController.Get() in D:\\Code\\VSProjects\\ApplicationInsightsModernLogger\\ApplicationInsightsModernLogger\\Controllers\\ValuesController.cs:line 27\r\n   at lambda_method(Closure , Object , Object[] )\r\n   at Microsoft.Extensions.Internal.ObjectMethodExecutor.Execute(Object target, Object[] parameters)\r\n   at Microsoft.AspNetCore.Mvc.Internal.ActionMethodExecutor.SyncObjectResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)\r\n   at Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker.InvokeActionMethodAsync()\r\n   at Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker.InvokeNextActionFilterAsync()\r\n   at Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker.Rethrow(ActionExecutedContext context)\r\n   at Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)\r\n   at Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker.InvokeInnerFilterAsync()\r\n   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeNextResourceFilter()\r\n   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.Rethrow(ResourceExecutedContext context)\r\n   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)\r\n   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeFilterPipelineAsync()\r\n   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeAsync()\r\n   at Microsoft.AspNetCore.Builder.RouterMiddleware.Invoke(HttpContext httpContext)\r\n   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)",
                "ConnectionId": "0HLJ42AH8F859"
            }
        }
    }
}```

File: src/ILogger/ApplicationInsightsLoggerExtensions.cs [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public static class ApplicationInsightsLoggerExtensions
+    {
+        /// <summary>
+        /// Adds an ApplicationInsights logger named 'ApplicationInsights' to the factory.
+        /// </summary>
+        /// <param name="builder">The <see cref="ILoggingBuilder"/> to use.</param>
+        public static ILoggingBuilder AddApplicationInsights(this ILoggingBuilder builder)
+        {
+            return builder.AddApplicationInsights((applicationInsightsOptions) => { });
```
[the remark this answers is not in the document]
RamjotSingh: ILoggingBuilder is supported in Console apps as well. I think @davidfowl mentioned that ILoggingBuilder is the new way to go.

Logging.sln
```diff
@@ -63,6 +63,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NLogTarget.NetCoreApp10.Tes
 EndProject
```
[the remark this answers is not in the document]
RamjotSingh: Once people agree that the approach I fine I will add unit tests.

src/ILogger/ILogger.csproj
```diff
@@ -0,0 +1,54 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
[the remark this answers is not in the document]
davidfowl: No let’s just target the 2.0 version please. It should be using the external scope provider and the new ILoggerBuilder extension methods. 

cc @pakrym can you jump in here and help with the design.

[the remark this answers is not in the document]
RamjotSingh: Yep it is a NetCore2.1 construct present 2.1 onwards.

src/ILogger/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,203 @@
...
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Implementation;
+
+    /// <summary>
+    /// Application insights logger implementation for <see cref="ILogger"/>.
+    /// </summary>
+    /// <seealso cref="Microsoft.Extensions.Logging.ILogger" />
+    public class ApplicationInsightsLogger : ILogger, IDisposable
```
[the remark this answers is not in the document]
RamjotSingh: I was looking at existing implementations (Console etc.) and saw those were public. Having this public gives people the freedom to have their own implementations (e.g. one that bypasses DI completely and creates an instance). So having this public should enable those scenarios.

src/ILogger/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,129 @@
...
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.Extensions.Options;
+
+    /// <summary>
+    /// Application insights logger provider.
+    /// </summary>
+    /// <seealso cref="Microsoft.Extensions.Logging.ILoggerProvider" />
+    /// <seealso cref="Microsoft.Extensions.Logging.ISupportExternalScope" />
+    [ProviderAlias("ApplicationInsights")]
+    public class ApplicationInsightsLoggerProvider : ILoggerProvider, ISupportExternalScope
```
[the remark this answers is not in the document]
RamjotSingh: Same as above.

I was looking at existing implementations (Console etc.) and saw those were public. Having this public gives people the freedom to have their own implementations (e.g. one that bypasses DI completely and creates an instance). So having this public should enable those scenarios.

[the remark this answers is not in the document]
pakrym: It is also useful for filtering. So you can do` builder.WithFilter<ApplicationInsightsLoggerProvider>(...)`

src/ILogger/ILogger.csproj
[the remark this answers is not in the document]
RamjotSingh: I guess if we are not going to remove the ILogger implementations in the existing ASPNetCore package (until 3.x?), then it is fine for us to target NetStandard2.0. We can in the ASPNetCore package for NetStandard 1.x users leave the ILogger implementation in place and then for NetStandard 2.0+ we can shift to using this implementation. Does that sound good @cijothomas ?

src/ILogger/ApplicationInsightsLoggerExtensions.cs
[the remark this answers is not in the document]
pakrym: `ILoggingFactory` extension methods are obsolete in 2.2 and removed in 3.0. We've added `LoggerFactory.Create` method in https://github.com/aspnet/Extensions/pull/640 to simplify `ILoggerBuilder` usage in non-hosted apps.

src/ILogger/ApplicationInsightsLogger.cs
RamjotSingh: Removing this for the initial release.

src/ILogger/ILogger.csproj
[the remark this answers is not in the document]
RamjotSingh: Talked offline to @cijothomas and he agreed that for now the plan looks fine. There are distribution logistics which we will need to figure out but they are orthogonal to this PR.

src/ILogger/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// <exception cref="System.ArgumentNullException">
+        /// telemetryConfiguration
+        /// or
+        /// loggingFilter
+        /// or
+        /// applicationInsightsLoggerOptions.
+        /// </exception>
+        public ApplicationInsightsLoggerProvider(
+            TelemetryConfiguration telemetryConfiguration,
+            IOptions<ApplicationInsightsLoggerOptions> applicationInsightsLoggerOptions)
```
pakrym: The current pattern of using `IOptions` doesn't react to configuration changes. Not sure how important it is for this implementation however.

RamjotSingh: The way AI is configured today there isn't really an option to keep shifting the pipeline. Hence I went with IOptions and not IOptionsMonitor. If this changes, we can look into using IOptionsMonitor.

RamjotSingh: Still @cijothomas  to confirm in case I overlooked a case.

src/ILogger/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,214 @@
...
+                        if (activeScope is IReadOnlyList<KeyValuePair<string, object>> activeScopeDictionary)
+                        {
+                            foreach (KeyValuePair<string, object> item in activeScopeDictionary)
+                            {
+                                dict[item.Key] = Convert.ToString(item.Value, CultureInfo.InvariantCulture);
+                            }
+                        }
+                        else
+                        {
+                            stringBuilder.Append(" => ").Append(activeScope);
```
davidfowl: This doesn’t avoid the closure allocation, use builder.

src/ILogger/ApplicationInsightsLoggerOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to track exceptions as <see cref="ExceptionTelemetry"/>.
+        /// </summary>
+        public bool TrackExceptionsAsExceptionTelemetry { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether EventId and EventName properties should be included in telemetry.
+        /// </summary>
+        public bool IncludeEventId { get; set; }
```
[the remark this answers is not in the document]
RamjotSingh: Flipped the IncludeEventId to ExcludeEventId and added an ExcludeScope.

By default we log them, but users can choose to skip them.

src/ILogger/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// The telemetry client to be used to log messages to Application Insights.
+        /// </summary>
+        private readonly TelemetryClient telemetryClient;
+
+        /// <summary>
+        /// The collection of application insights loggers stored against categoryName.
+        /// categoryName -> <see cref="ApplicationInsightsLogger"/> map.
+        /// </summary>
+        private readonly ConcurrentDictionary<string, ApplicationInsightsLogger> applicationInsightsLoggers
+            = new ConcurrentDictionary<string, ApplicationInsightsLogger>();
```
davidfowl: Why do you need this at all?

RamjotSingh: One option is to let Provider create a new logger every time and use that (Transient) pattern. Or hold on to instances in general and keep reusing them. I looked at [ConsoleLogger](https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs) and saw the pattern of holding on to instances and followed the same pattern here.

There is an additional assignement of SDK version is Logger as well, however that can be pulled in Provider.

src/ILogger/ApplicationInsightsLoggerOptions.cs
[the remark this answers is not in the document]
RamjotSingh: Aligned with ConsoleLoggerOptions. EventId is always logged, Scopes has a switch called IncludeScope default set to false.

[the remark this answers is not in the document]
pakrym: Don't set it to `false` here, in console logging we are saving screen space, for structured loggers scopes should be logged by default.

src/ILogger/ApplicationInsightsLoggerProvider.cs
pakrym: If it wasn't for option reload we wouldn't hold onto loggers in console logger either. `LoggerFactory` would cache `ILogger` instances for you

pakrym: And `SetScopeProvider` is called before any calls to `CreateLogger` so you are safe there too.

RamjotSingh: Hmm, alright in that case I will remove caching logic inside the SDK.

src/ILogger/ApplicationInsightsLoggerOptions.cs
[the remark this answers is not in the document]
RamjotSingh: Done. Scopes are now included by default.

src/ILogger/ApplicationInsightsLoggerProvider.cs
cijothomas: Yes IOptions works fine for now. (we can enhance later, but not required for this PR)

RamjotSingh: Resolving this for now then.

src/ILogger/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,120 @@
...
```
davidfowl: Did you forget to remove this?

src/ILogger/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        public static ILoggingBuilder AddApplicationInsights(
+            this ILoggingBuilder builder,
+            Action<ApplicationInsightsLoggerOptions> configureApplicationInsightsOptions)
+        {
+            if (configureApplicationInsightsOptions == null)
+            {
+                throw new ArgumentNullException(nameof(configureApplicationInsightsOptions));
+            }
+
+            // This ensures that we don't override the TelemetryClient if one was added using .AddApplicationInsightsTelemetry()
```
davidfowl: Who’s adding the telemetry client?

src/ILogger/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,120 @@
...
+            IOptions<ApplicationInsightsLoggerOptions> applicationInsightsLoggerOptions)
+        {
+            if (telemetryConfiguration == null)
+            {
+                throw new ArgumentNullException(nameof(telemetryConfiguration));
+            }
+
+            this.applicationInsightsLoggerOptions = applicationInsightsLoggerOptions?.Value ?? throw new ArgumentNullException(nameof(applicationInsightsLoggerOptions));
+
+            this.telemetryClient = new TelemetryClient(telemetryConfiguration);
```
davidfowl: Ah I see. What happens if there are 2 of these in the system. Is there still a single collector instance shipping logs to an http backend?

RamjotSingh: No in that case 2 telemetry clients will individually ship logs to AI. The equation however is not always this simple and there is a whole world of TelemetrySinks which can make everything sync or Async ordered etc. (I have recently started exploring the depths of AI SDK so my knowledge is still limited).

So in short if you are using everything default, TelemetryClients will be shipping their individual entries to AI opening 2 connections. This is effectively no different than 2 micro-services using the same AI key running on same machine.

src/ILogger/ApplicationInsightsLoggingBuilderExtensions.cs
RamjotSingh: No one is adding TelemetryClient per say. TelemetryConfiguration is what is added using the existing ASPNetCore extension methods (in ASPNetCore package). However all of this is true provided the user sets up everything correctly. There are chances AI pipeline might not be set correctly (developers might choose to not have it running on your local machine or something similar). For such cases, I am adding a dummy Configuration which is used to create a TelemtryClient which runs the full AI pipeline, but does not send the telemetry outside of the machine.

LoggerProvider takes TelemetryConfiguration and creates TelemetryClient from it and then passes this to every Logger instance.

There is a rather interesting side case as well. ILogger instance can be asked for during the Startup.cs call. However at this point if during Webhost setup, developer did not call .UseApplicationInsights method, all logs emiited during Startup.cs are lost since AI key was not yet setup. Additionally since .UseApplicationInsights method does not allow for adding additional sinks, telemetryInitializers or processors telemetry might be a lot lacking. However once the ConfigureServices and Configure call is made the full pipeline will be created and the second instance of ApplicationInsightsLoggerProvider will be shipping logs to AI and running full pipelines.

But all of this is existing and this code and not change any of that. The console sample at the very top gives a brief idea of how the pipeline comes into effect.

src/ILogger/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    using System;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.DependencyInjection.Extensions;
+    using Microsoft.Extensions.Logging.ApplicationInsights;
+
+    /// <summary>
+    /// Extensions methods to add and configure application insights logger.
+    /// </summary>
+    public static class ApplicationInsightsLoggingBuilderExtensions
```
davidfowl: The instrumentation key should be a parameter here’s the caller shouldn’t have to drop down to service collection shenanigans to configure it.

RamjotSingh: Pulling Instrumentation key alone is not enough. TelemetrySinks, TelemetryChannels, TelemetryInitializers and TelemetryProcessors would still need to be setup. However these are not always required to be setup.

One option is to have an extension method which takes instrumentation key to get people up and running while rest of the stuff can be added later using .Configure<TelemetryConfiguration> if needed.

src/ILogger/ApplicationInsightsLoggerProvider.cs
davidfowl: So there will be 2 background threads in ASP.NET Core applications that use this logger and enable the AI site extension right?

RamjotSingh: Yes, 2 telemetry clients exist and these 2 telemetry clients are used in different scopes. One if used in the scope of startup.cs while the other one is used for the rest of the pipelines (requests etc.). These can emit the logs to (potentially) AI instances. This is however not a feature, but a side effect.

davidfowl:
I'm hope it would look something like this:

```C#
var services = new ServiceCollection();            
services.AddLogging(loggingBuilder =>
{
    loggingBuilder.AddApplicationInsights("--YourAIKeyHere--");
});

var serviceProvider = services.BuildServiceProvider();
var loggerFactory = serviceProvider.GetRequiredService<ILoggerFactory>();
var logger = loggerFactory.CreateLogger("MyCategory");

using (logger.BeginScope("Test"))
{
    logger.LogInformation("Logger is working");
}
```

in 3.0 it'll look like this:

```C#
var loggerFactory = LoggerFactory.Create(builder => 
{
    builder.AddApplicationInsights("--YourAIKeyHere--");
});

using (logger.BeginScope("Test"))
{
    logger.LogInformation("Logger is working");
}
```

RamjotSingh:
As @pakrym pointed out, right now Logging and ASPNetCore integrations are stepping on each other's toes. This is partially because ASPNetCore package contains things which are not ASPNetCore specific but in general NetCore constructs (things like IServiceCollection bits etc.). 

Before AI 3.X SDK is shipped, it might make sense to shuffle packages around a bit and have a core package which caters to NetCore way of things. But until then here is what I propose.

Let's figure out what is the bare minimum when we talk about logging. At this point that bare minimum is nothing because I am adding a dummy TelemetryClient to the container to ensure things are not broken if developer forgets to setup AI.

We can pull the instrumentation key and raise the bare minimum (this satisfies @davidfowl 's ask). This splits the use case like this

#### AspNetCore application
- Basic user - Just set instrumentation key for logger and everything works out of the box.
- Advanced user - Use the .AddApplicationInsightsTelemetry extension methods to customize the pipeline.

#### NetStandard2.X based application
- Basic user - Same as above
- Advanced user - use 
```.Configure<TelemetryConfiguration>``` to configure TelemetryConfiguration, followed by 
```cs
services.AddSingleton<TelemetryClient>(new TelemetryClient(provider.GetRequiredService<IOptions<TelemetryConfiguration>>))
```

This still means that touching logging would impact rest of the AspNetCore pipeline and vice versa but that is something we can debate and see if we want to continue with this or change it.

davidfowl:
@RamjotSingh lets file an issue for @pakrym 's suggestion and clean up the API to add the InstrumentationKey

RamjotSingh:
Logged the issue https://github.com/Microsoft/ApplicationInsights-dotnet-logging/issues/241.
Additionally I shifted to using IOptions<TelemetryConfiguration> instead of TelemetryConfiguration which means the current package will behave nicely with the existing ecosystem even if we were to add a function here to configure TelemetryConfiguration. But I am intentionally leaving this out of this PR so that we don't accidently lock ourselves into a corner.

RamjotSingh:
@cijothomas, @davidfowl , @pakrym  Can you folks give this one final look. I think this is ready for an initial release if there is no additional feedback. I have not added the extension method to specify instrumentation key in place at this point. However with the latest iteration it is possible to do it without messing up things. So if we feel we need it right now I can add it very easily.

src/ILogger/ILogger.csproj
```diff
@@ -8,6 +8,7 @@
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
     <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
     <RunCodeAnalysis>false</RunCodeAnalysis>
+    <DebugType>Full</DebugType>
```
davidfowl: What’s going on here?

RamjotSingh: Removed that line. Was trying to find the paths not tested by unit tests. I forgot that newer SDK had the fix for unit tests.

davidfowl:
We need to instrumentation key specified directly in the method call to AddApplicationInsights.

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
```diff
@@ -0,0 +1,236 @@
...
+            Assert.AreEqual(" => TestScope", (itemsReceived[0] as ISupportProperties).Properties["Scope"]);
+            Assert.AreEqual("Value", (itemsReceived[0] as ISupportProperties).Properties["Key"]);
+
+            Assert.AreEqual(" => TestScope", (itemsReceived[1] as ISupportProperties).Properties["Scope"]);
+            Assert.AreEqual("Value", (itemsReceived[1] as ISupportProperties).Properties["Key"]);
+
+            itemsReceived.Clear();
+
+            // Case where Scope is NOT Included
+            serviceProvider = ILoggerIntegrationTests.SetupApplicationInsightsLoggerIntegration(
```
pakrym: Split this test and `ApplicationInsightsLoggerHonorsLogExceptionsAsExceptionTelemetrySwitch` into two.

RamjotSingh:
@davidfowl I have added a new extension method which takes an instrumentation key. The newer sample looks identical to the one you put in above. I have not put in a sample for 3.0 but it should look similar to yours as well.

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
RamjotSingh: Done.

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
```diff
@@ -0,0 +1,277 @@
...
+
+            // Case where Scope is NOT Included
+            IServiceProvider serviceProvider = ILoggerIntegrationTests.SetupApplicationInsightsLoggerIntegration(
+                (telemetryItem, telemetryProcessor) => itemsReceived.Add(telemetryItem),
+                configureTelemetryConfiguration: null,
+                configureApplicationInsightsOptions: (appInsightsLoggerOptions) => appInsightsLoggerOptions.IncludeScopes = false);
+
+            ILogger<ILoggerIntegrationTests> testLogger = serviceProvider.GetRequiredService<ILogger<ILoggerIntegrationTests>>();
+
+            testLogger.BeginScope("TestScope");
```
davidfowl: nit add usings

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
```diff
@@ -0,0 +1,277 @@
...
+
+            testLogger.LogInformation("Testing");
+            testLogger.LogError(new Exception("TestException"), "Exception");
+            testLogger.LogInformation(new EventId(100, "TestEvent"), "TestingEvent");
+            testLogger.LogCritical("Critical");
+            testLogger.LogTrace("Trace");
+            testLogger.LogWarning("Warning");
+            testLogger.LogDebug("Debug");
+
+            Assert.AreEqual(7, itemsReceived.Count);
```
davidfowl: Shouldn't it assert the content and levels?

src/ILogger/ILogger.csproj
```diff
@@ -0,0 +1,54 @@
...
+  <PropertyGroup>
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+    <IncludeSymbols>True</IncludeSymbols>
+    <PackageId>Microsoft.Extensions.Logging.ApplicationInsights</PackageId>
+    <Title>Microsoft Logging Extensions for ApplicationInsights</Title>
+    <Description>Application Insights ILogger allows forwarding events from ILogger to Application Insights. Application Insights will collect your logs from multiple sources and provide rich powerful search capabilities. Privacy statement: https://go.microsoft.com/fwlink/?LinkId=512156</Description>
+    <PackageTags>$(PackageTags) ILogger ILoggerBuilder ILoggerProvider</PackageTags>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)'=='Release'">
```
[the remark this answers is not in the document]
RamjotSingh: Removed DebugType=Full and added sourcelink.

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
RamjotSingh: Done.

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
RamjotSingh: Done.

pakrym:
LGTM

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
```diff
@@ -0,0 +1,296 @@
...
+                using (testLogger.BeginScope<IReadOnlyCollection<KeyValuePair<string, object>>>(new Dictionary<string, object> { { "Key", "Value" } }))
+                {
+                    testLogger.LogInformation("Testing");
+                    testLogger.LogError(new Exception("TestException"), "Exception");
+                }
+            }
+
+            Assert.AreEqual(" => TestScope", (itemsReceived[0] as ISupportProperties).Properties["Scope"]);
+            Assert.AreEqual("Value", (itemsReceived[0] as ISupportProperties).Properties["Key"]);
+
```
cijothomas: lets check the actual message in the itemReceived as well.

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
```diff
@@ -0,0 +1,296 @@
...
+                using (testLogger.BeginScope<IReadOnlyCollection<KeyValuePair<string, object>>>(new Dictionary<string, object> { { "Key", "Value" } }))
+                {
+                    testLogger.LogInformation("Testing");
+                    testLogger.LogError(new Exception("TestException"), "Exception");
+                }
+            }
+
+            Assert.IsFalse((itemsReceived[0] as ISupportProperties).Properties.ContainsKey("Scope"));
+            Assert.IsFalse((itemsReceived[0] as ISupportProperties).Properties.ContainsKey("Key"));
+
```
cijothomas: lets check the actual message in the itemReceived as well. and the count is also as expected.

test/ILogger.NetStandard.Tests/ILoggerIntegrationTests.cs
```diff
@@ -0,0 +1,296 @@
...
+            Assert.AreEqual("Value", (itemsReceived[1] as ISupportProperties).Properties["Key"]);
+        }
+
+        /// <summary>
+        /// Ensures that the <see cref="ApplicationInsightsLoggerOptions.IncludeScopes"/> switch is honored and scopes are excluded
+        /// when switch is false.
+        /// </summary>
+        [TestMethod]
+        [TestCategory("ILogger")]
+        public void ApplicationInsightsLoggerAddsScopeWhenSwitchIsFalse()
```
cijothomas: AddsScope =>DoesNotAddScope - minor renaming

src/ILogger/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// Begins a logical operation scope.
+        /// </summary>
+        /// <typeparam name="TState">Current state.</typeparam>
+        /// <param name="state">The identifier for the scope.</param>
+        /// <returns>
+        /// An IDisposable that ends the logical operation scope on dispose.
+        /// </returns>
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            return this.ExternalScopeProvider != null ? this.ExternalScopeProvider.Push(state) : NullScope.Instance;
```
cijothomas: We need to do this only if (this.ExternalScopeProvider != null)

src/ILogger/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <summary>
+        /// Checks if the given <paramref name="logLevel" /> is enabled.
+        /// </summary>
+        /// <param name="logLevel">level to be checked.</param>
+        /// <returns>
+        ///   <c>true</c> if enabled.
+        /// </returns>
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return this.telemetryClient.IsEnabled();
```
cijothomas: This does not look correct - we need to check loglevel to make this decision. Also missing are unit tests to validate that we respect log levels.

RamjotSingh:
@cijothomas Here is how someone can configure ApplicationInsights to only log above a specific level.

```cs
               loggingBuilder.AddFilter((logLevel) =>
                {
                    if (logLevel > LogLevel.Information)
                    {
                        return true;
                    }
                    else
                    {
                        return false;
                    }
                })
```

I think this is what @pakrym talked about when referring to Filtering.

src/ILogger/ILogger.csproj
```diff
@@ -0,0 +1,64 @@
...
+  <PropertyGroup>
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+    <IncludeSymbols>True</IncludeSymbols>
+    <PackageId>Microsoft.Extensions.Logging.ApplicationInsights</PackageId>
+    <Title>Microsoft Logging Extensions for ApplicationInsights</Title>
+    <Description>Application Insights ILogger allows forwarding events from ILogger to Application Insights. Application Insights will collect your logs from multiple sources and provide rich powerful search capabilities. Privacy statement: https://go.microsoft.com/fwlink/?LinkId=512156</Description>
+    <PackageTags>$(PackageTags) ILogger ILoggerBuilder ILoggerProvider</PackageTags>
+  </PropertyGroup>
+  
+
```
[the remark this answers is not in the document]
RamjotSingh: I actually removed it on @davidfowl 's recommendation. Adding it back for now as I guess this is needed.

src/ILogger/ApplicationInsightsLogger.cs
RamjotSingh: Method returns an IDisposable regardless so I need to return a dummy object.

src/ILogger/ApplicationInsightsLogger.cs
RamjotSingh: Filtering is taken care of my LoggingFactory so we are Provider and Logger level don't need to care about it.

```cs
               loggingBuilder.AddFilter((logLevel) =>
                {
                    if (logLevel > LogLevel.Information)
                    {
                        return true;
                    }
                    else
                    {
                        return false;
                    }
                }).AddApplicationInsights();
```

davidfowl:
Yes filtering is built into the factory as a first class citizen.

lmolkova:
@RamjotSingh  we fixed build issue (https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/243) and are going to ship it in next beta release. Would you mind re-enabling analyzers nugets as a separate PR?

lmolkova:
@RamjotSingh thanks a lot for the contribution! 🥇

PR #243, opened 2018-12-31
Fix fxcop running on netstandard2.0 and failing

lmolkova:
In https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/239 new ILogger was introduced that targets netstandard 2.0 and build fails because of code analyzers 

```
CA0055 : Could not identify platform for 'E:\A\_work\107\bin\Release\src\ILogger\netstandard2.0\Microsoft.Extensions.Logging.ApplicationInsights.dll'.
CA0052 : No targets were selected
```

https://stackoverflow.com/questions/46617261/enabling-code-analysis-for-net-core-2-0-project-results-ca0055-and-ca0052-error

> This error is caused by using an old version of Code Analysis with .NET Core. This old version is only for non-.NET Core applications.
The solution is to disable the old Code Analysis for .NET Core projects and install the new version of Code Analysis, which is now a NuGet package. (The reason you probably want to disable the old Code Analysis tool for your project and NOT uninstall it is so that you can still use the old Code Analysis with old .NET applications, such as .NET 4.5.)

To fix it, we need to remove `RunCodeAnalysis` from Common props and set it on the projects that need it.

### DIFF
--- a/Common.props
+++ b/Common.props
@@ -19,7 +19,6 @@
   <PropertyGroup Condition="'$(Configuration)' == 'Release'">
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
-    <RunCodeAnalysis>true</RunCodeAnalysis>
     <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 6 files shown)


MS-TimothyMothra:
Hey! I've seen this before :)

https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/0ba3e48110cbb200a6f72091f7b0b003e31f6532/Common.props#L37-L43

PR #245, opened 2018-12-31
Update changelog and re-enable analyzers on ilogger

lmolkova:
Follow up on #239 update changelog and enable analyzers


lmolkova:
@cijothomas  please review

PR #282, opened 2019-04-26
ILogger - Exception.Message is populated as ExceptionTelemetry.Message, Bump version

cijothomas:
ILogger - If an exception is passed to log, then Exception.Message is populated as ExceptionTelemetry.Message. 

If TrackExceptionsAsExceptionTelemetry is false, then Exception.Message is stored as custom property "ExceptionMessage"

Foe example, consider the following log call.
testLogger.LogError(new Exception("ExceptionMessage"), "LoggerMessage");
The ExceptionTelemetry created will contain "ExceptionMessage" in its Message field. "LoggerMessage" will be in custom property "FormattedMessage"

If TrackExceptionsAsExceptionTelemetry is false, then TraceTelemetry is created, and "LoggerMessage" will be in its Message field. "ExceptionMessage" will be in custom property "ExceptionMessage"

Prior to this, when an exception is passed to log, its Exception.Message was not stored anywhere.

Also bumped and updated versions.

### DIFF
--- a/src/ILogger/ApplicationInsightsLogger.cs
+++ b/src/ILogger/ApplicationInsightsLogger.cs
@@ -86,13 +86,18 @@ namespace Microsoft.Extensions.Logging.ApplicationInsights
                         formatter(state, exception),
                         ApplicationInsightsLogger.GetSeverityLevel(logLevel));
                     this.PopulateTelemetry(traceTelemetry, state, eventId);
+                    if (exception != null)
+                    {
+                        traceTelemetry.Properties.Add("ExceptionMessage", exception.Message);
+                    }
+
                     this.telemetryClient.TrackTrace(traceTelemetry);
                 }
                 else
                 {
                     ExceptionTelemetry exceptionTelemetry = new ExceptionTelemetry(exception)
                     {
-                        Message = formatter(state, exception),
+                        Message = exception.Message,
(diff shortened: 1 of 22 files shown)


RamjotSingh:
How does this behave when users use LogError with exception object in it? I think we might be regressing that here.

cijothomas:
@RamjotSingh i have edited to give example. (I submitted PR with empty description, sorry)

src/ILogger/ApplicationInsightsLogger.cs
```diff
@@ -86,16 +86,22 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+                    if (exception != null)
+                    {
+                        traceTelemetry.Properties.Add("ExceptionMessage", exception.Message);
+                    }
+
                     this.telemetryClient.TrackTrace(traceTelemetry);
                 }
                 else
                 {
                     ExceptionTelemetry exceptionTelemetry = new ExceptionTelemetry(exception)
```
lmolkova: this branch seems to be executed when exception == null  && this.applicationInsightsLoggerOptions.TrackExceptionsAsExceptionTelemetry (else), but it does not seem right to track exception and  Message = exception.Message will throw. Am I missing something?

cijothomas: discussed offline. this is not the case.

PR #286, opened 2019-05-17
update analyzers and resolve new issues

MS-TimothyMothra:


src/EtwCollector/EtwTelemetryModule.cs
```diff
@@ -180,7 +180,7 @@ private void EnableProviders()
             {
                 ProviderGuid = TplActivities.TplEventSourceGuid,
                 Level = TraceEventLevel.Always,
-                Keywords = TplActivities.TaskFlowActivityIdsKeyword
+                Keywords = TplActivities.TaskFlowActivityIdsKeyword,
```
cijothomas: is this a compilation error?

MS-TimothyMothra: not a complication error. seems like a code standards thing, they want to include the trailing comma  everywhere. 
i thought it's a very minor non-functional change so i did it everywhere.

PR #291, opened 2019-05-24
sdl fix system.net.http

MS-TimothyMothra:
taking an explicit dependency on System.Net.Http will override any implicit dependencies.

I'm trying to copy @lmolkova's solution from here: https://github.com/microsoft/ApplicationInsights-dotnet/pull/1147

Confirmed that this fixes the security scan.

### DIFF
--- a/src/DiagnosticSourceListener/DiagnosticSourceListener.csproj
+++ b/src/DiagnosticSourceListener/DiagnosticSourceListener.csproj
@@ -52,6 +52,7 @@
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.5.0" />
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
   </ItemGroup>
 
   <Import Project="..\CommonShared\CommonShared.projitems" Label="Shared" />
--- a/src/EventSourceListener/EventSourceListener.csproj
+++ b/src/EventSourceListener/EventSourceListener.csproj
@@ -52,6 +52,7 @@
     <PackageReference Include="Microbuild.Core" Version="0.3.0">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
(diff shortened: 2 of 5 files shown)


src/DiagnosticSourceListener/DiagnosticSourceListener.csproj
```diff
@@ -52,6 +52,7 @@
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.5.0" />
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
```
lmolkova: this is only applicable to netstandard (<2.0) targets and is bad for netfx targets
could you please make sure for all projects we ally it to  netstandard1.3 and 1.6 only?

lmolkova:
I *think* standard SDL check is not clever enough. So what happens is:
-  our lib depends on .NET Core library 1.6.1 and does not depend on Http
- .NET Core lib depends on System.Net.Http >= 4.3.0
- If app needs http, it's app responsibility to install a non-vulnerable version.

now this SDL check forces us to add a dependency to the library we don't use or need in logging. I'm not sure if we can do anything here like is SDL check has some help or owner that can help to clarify if this is the right approach?

src/DiagnosticSourceListener/DiagnosticSourceListener.csproj
MS-TimothyMothra: Thank you. I will do this

MS-TimothyMothra:
Thank you for the explanation. 
Let me verify that this resolve the alerting rule for all repos.
Then I can re-review the guidance or reach out to decide if this is the "best" solution. :)

lmolkova:
By the way, next base 2.11-beta1 will force system.net.htpp to 4.3.4 and this PR is likely not needed

MS-TimothyMothra:
I created a build of latest Base SDK and verified that it fixes the issues in Logging SDK. It does! :)
I think we can close this issue. 
Thanks @lmolkova

